Issue #28, opened 2021-06-11
API Discovery


jmprusi:
I'm working on implementing this concept as a PoC, but I would like to get some feedback! 

When adding a new application that contains an API, to reduce the number of objects that a user has to author,
we can leverage annotations. A good place to annotate could be the Service used to drive traffic to the application 
PODs.

You will recognize some of those from the "3scale discovery spec". 

### Annotations:

- **discovery.kuadrant.io/scheme**: *OPTIONAL* Either HTTP or HTTPS specifies how the kuadrant gateway will connect to this API.
- **discovery.kuadrant.io/api-name**: *OPTIONAL* If not set, the name of the API can be matched with the service name.
- **discovery.kuadrant.io/tag**: OPTIONAL A special tag used to distinguish this deployment between several instances of the API.
- **discovery.kuadrant.io/oas-path**:  Path where the OAS of the service is exposed. Kuadrant will contact the service and try to retrieve the OAS spec.
- **discovery.kuadrant.io/oas-configmap**: Configmap that contains the OAS spec.
- **discovery.kuadrant.io/port**: Only required if there are multiple ports in the service. Either the Name of the port or the Number.

Labels:
- **discovery.kuadrant.io/enabled:**: *REQUIRED* true or false, marks the object to be discovered by kuadrant.

Complete annotated Service example:

```yaml
apiVersion: v1
kind: Service
metadata:
  name: dogs-api
  annotations:
    discovery.kuadrant.io/scheme: "https"
    discovery.kuadrant.io/api-name: "dogs-api"
    discovery.kuadrant.io/tag: "production"
    discovery.kuadrant.io/oas-path: "/api/openapi/json"
  labels:
    discovery.kuadrant.io/enabled: "true"
spec:
  selector:
    app: dogs-api
  ports:
    - name: HTTP
      protocol: TCP
      port: 80
      targetPort: 9376
```

In case that the API is hosted outside the cluster, the developer can use the ExternalName service type, this way we keep only having to reconcile Services and not extra types. An example:

```yaml
apiVersion: v1
kind: Service
metadata:
  name: dogs-api
  annotations:
    discovery.kuadrant.io/scheme: "https"
    discovery.kuadrant.io/api-name: "dogs-api"
    discovery.kuadrant.io/tag: "production"
    discovery.kuadrant.io/oas-path: "/api/openapi/json"
  labels:
    discovery.kuadrant.io/enabled: "true"
spec:
  type: ExternalName
[...]

PR #29, opened 2021-06-17
Service discovery implementation

jmprusi:
Just trying to get it working and figuring out new issues/edge cases. 

Example annotations (https://github.com/Kuadrant/kuadrant-controller/issues/28):

```yaml
discovery.kuadrant.io/scheme: "https"
discovery.kuadrant.io/api-name: "dogs-api"
discovery.kuadrant.io/tag: "production"
discovery.kuadrant.io/port: 80
discovery.kuadrant.io/oas-configmap: configmapName
```

To try it right now:

```bash
make local-setup
kubectl apply -f ./samples
kubectl port-forward --namespace kuadrant-system deployment/kuadrant-gateway 8080:8080 8443:8443 &
curl -H "Authorization: APIKEY JUSTFORDEMOSOBVIOUSLYqDQsqSPMHkRhriEOtcRx" -H "Host: api.animaltoys.127.0.0.1.nip.io:8080" localhost:8080/cats/toys -v
``` 

TODO:

- [ ]  Reconcile Configmap
- [ ] Fix some broken reconciliation
- [ ] Add new tests and fix current ones
- [ ] Some quick refactoring and cleaning


controllers/discovery/service_controller.go
```diff
@@ -0,0 +1,245 @@
...
+				enabled, _ := strconv.ParseBool(val)
+				return enabled
+			}
+			return false
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			// In case the update object had the kuadrant label set to true, we need to reconcile it.
+			if val, ok := e.ObjectOld.GetLabels()[kuadrantDiscoveryLabel]; ok {
+				enabled, _ := strconv.ParseBool(val)
+				return enabled
```
eguzki, line 229: If I remove the discovery label, the event is accepted. Should not be discarded only based on the new state?

jmprusi: If you remove the `discovery label`, it means that the `discovery label` existed before, we need to handle that situation somehow, for example, removing the API object created by that service.

samples/api1.yaml
```diff
@@ -1,14 +1,61 @@
...
+    - port: 80
+      protocol: TCP
+      targetPort: 3000
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: cats-oas
+data:
+  openapi.yaml: |
```
eguzki, line 51: OpenAPI 3.0 is not valid. I spotted two issues:

- required field missing: `info.version`
- required (responses](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.2.md#responsesObject) field missing in the operation object.

Minim valid OAS3:
```
  openapi: "3.0.0"                   
 info:                              
   title: "toy API"                 
   version: "1.0.0"                 
 servers:                           
   - url: http://toys/              
 paths:                             
   /toys:                           
     get:                           
       operationId: "getToys"       
       responses:                   
         200:                       
           description: "success"   
```

consider adding validation call after parsing is done.`

apis/networking/v1beta1/api_types.go
```diff
@@ -46,17 +46,17 @@ type SecurityRequirements []SecurityRequirement
...
+type APIDefinition struct {
+	OAS string `json:"oas"`
+}
+
+type Tag struct {
+	Name        string      `json:"name"`
+	Destination Destination `json:"destination"`
 }
 
 type SecurityScheme struct {
```
eguzki, line 62: Many objects like this belong to the APIProduct type.

apis/networking/v1beta1/apiproduct_types.go
```diff
@@ -60,35 +60,22 @@ type ProductInformation struct {
...
 }
 
-type BackendServer struct {
-	API         string      `json:"API"`
-	Destination Destination `json:"destination"`
+type OpenIDConnectAuthCredentials struct {
+	Endpoint string `json:"endpoint"`
 }
 
 type Destination struct {
```
eguzki, line 76: `Destination` object type belong to the API object type.

Good to have the destination in the API, instead of the APIProduct

eguzki:
I like the approach of annotating the service.

IMO, we can skip (for now) the reconciliation of the configmap holding the OAS. If the user wants to update the OAS, the user can always create a new config map with the updated OAS content and update the service annotations. The controller will update the API internally.

There are other issues, but overall looks good and we can merge. One TODO I have not seen is the TODO to clean-up kuadrant objects and isitio resources when the service removes or sets to false the kuadrant label, or deletes the service.

eguzki:
Consider adding owner references to managed objects. 

`apis.networking.kuadrant.io` owned by Services
`virtualservices.networking.istio.io` owned by `apiproducts.networking.kuadrant.io`
`services.config.authorino.3scale.net` owned by by `apiproducts.networking.kuadrant.io`

The, the controller can add watches to theses object types easily to ensure the desired state is right. It also can be very useful to have a meaningful `status` field in the apiproducts with info about the managed resources

jmprusi:
> Consider adding owner references to managed objects.
> 
> `apis.networking.kuadrant.io` owned by Services
> `virtualservices.networking.istio.io` owned by `apiproducts.networking.kuadrant.io`
> `services.config.authorino.3scale.net` owned by by `apiproducts.networking.kuadrant.io`
> 
> The, the controller can add watches to theses object types easily to ensure the desired state is right. It also can be very useful to have a meaningful `status` field in the apiproducts with info about the managed resources

Yes, thank you, the owner refs are still missing in some places. I will review it and add/fix them during this week, but we need to take into account where those objects are created, if they are in different namespaces, an owner ref will not help there.

eguzki:
> > Consider adding owner references to managed objects.
> > `apis.networking.kuadrant.io` owned by Services
> > `virtualservices.networking.istio.io` owned by `apiproducts.networking.kuadrant.io`
> > `services.config.authorino.3scale.net` owned by by `apiproducts.networking.kuadrant.io`
> > The, the controller can add watches to theses object types easily to ensure the desired state is right. It also can be very useful to have a meaningful `status` field in the apiproducts with info about the managed resources
> 
> Yes, thank you, the owner refs are still missing in some places. I will review it and add/fix them during this week, but we need to take into account where those objects are created, if they are in different namespaces, an owner ref will not help there.

This is a good point. This is a new way of thinking, as I have been always working with namespaced operators. So, is it feasible (and wanted) to create API objects in the namespace where the service exists? Same questions for istio VirtualServices and authorino services when apiproduct is created in namespace A. Do they need to be created in `kuadrant-system` namespace?  

Ownership allows two features: a) builtin delete cascade b) builtin filter for watching events on resources. If ownership is not feasible, the controller can implement smart event filtering based on some labels the controller adds to the resources managed.

jmprusi:
> This is a good point. This is a new way of thinking, as I have been always working with namespaced operators. So, is it feasible (and wanted) to create API objects in the namespace where the service exists? Same questions for istio VirtualServices and authorino services when apiproduct is created in namespace A. Do they need to be created in `kuadrant-system` namespace?

I'm trying to avoid polluting several namespaces with "internal" objects of kuadrant. Also, as an `APIproduct` can point to multiple `APIs` in different namespaces, I prefer to have the ingress/authorization configuration centralized into "our" namespace. 

> Ownership allows two features: a) builtin delete cascade b) builtin filter for watching events on resources. If ownership is not feasible, the controller can implement smart event filtering based on some labels the controller adds to the resources managed.

Thanks! I (mostly) understand how kubernetes handles ownership/dependants and garbage collection, and for example, we are using some event filtering in this PR to only watch for labeled services. 

If this way of handling the objects proves too intrusive or becomes unmanageable, let's change it. But let's keep that discussion for another PR! :)

eguzki:
Ok, then the proposed solution is:

A)  API **internal** kuadrant objects in the same namespace as the customer service. I think it makes sense.
B) APIProduct kuadrant objects in any namespace (not limited to `kuadrant-system`) and has reference to API's in same or another namespaces. So there can be a APIProduct in namespace A having references to API's in namespace C and D.
C)  ingress/authorization configuration resourcs in `kuadrant-system` namespace. 

LGTM

So the owner reference of the Service in the kuadrant API resource makes sense and it is missing. 

I think we need to find a way to relate ingress/authorization configuration resources with the API Product, so the controller receives events when those resources change. Same thing for the API objects with API Product. Labels or annotations so when API changes, APIProduct reconciler gets an event (implemented in some filtermap).

eguzki:
Regarding `API tags`. I am not sure I understand them. API has an array of tags. But the source service can only be tagged with one tag. And it seems to be a 1:1 relation between API and Service. So, what is the idea behind having a array of tags? The OAS field in the API CRD is unique, but it may have several tags.

jmprusi:
> Regarding `API tags`. I am not sure I understand them. API has an array of tags. But the source service can only be tagged with one tag. And it seems to be a 1:1 relation between API and Service. So, what is the idea behind having a array of tags? The OAS field in the API CRD is unique, but it may have several tags.

Yes, I'm not sure of that object. The idea was to register multiple instances of an API (tags) as a single object, but we need to make sure that those APIs OAS are the same, etc.. We can change that objects if it doesn't make sense.

eguzki:
The API tags is also a good example about requirements from the customer. Like the lifecycle management. 

We need to provide an easy and straightforward way for the customer to test configuration (some APIProduct and OAS) in a "staging" env before promoting to "production".

jmprusi:
> The API tags is also a good example about requirements from the customer. Like the lifecycle management.
> 
> We need to provide an easy and straightforward way for the customer to test configuration (some APIProduct and OAS) in a "staging" env before promoting to "production".

do you mean that it makes sense? or that we should wait for an actual user requirement?

Changed the API object, now TAGs contain the API Definition too.

eguzki (approved):

PR #33, opened 2021-07-01
Fix integration tests

eguzki:


controllers/networking/helper_test.go
```diff
@@ -150,8 +150,8 @@ func CheckForDeploymentsReady(ns string, k8sClient client.Client) error {
 		return err
 	}
 
-	if len(deploymentList.Items) < 3 {
-		return errors.New("Expecting at least 3 Deployments")
+	if len(deploymentList.Items) < 4 {
```
jmprusi, line 153: we will need to change this function, at least send the number of deployments or a list of deployment names that need to be ready.

throw a TODO in there for now.

jmprusi (approved):

controllers/networking/helper_test.go
eguzki: Ok, I will add in other PR

PR #37, opened 2021-07-05
Apiproduct status conditions

eguzki:
APIProduct status based on often used pattern "conditions".

The new `ready` status would look like this:
```yaml
status:
    conditions:
    - lastTransitionTime: "2021-07-05T13:34:34Z"
      message: Ready
      reason: Ready
      status: "True"
      type: Ready
    observedgen: 8
```

Added a new status check to be reported in the status. It should be a good example to use for other reports in the status. The implemented status reporter checks that resources referenced in `spec.APIs` exist and the `mapping.prefix` is not duplicated.


pkg/common/k8s_utils.go
```diff
@@ -17,13 +17,17 @@ limitations under the License.
...
 	"fmt"
+	"sort"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
-	DeleteTagAnnotation = "kuadrant.io//delete"
+	DeleteTagAnnotation      = "kuadrant.io//delete"
```
jmprusi, line 29: Was that "//" intended, or is it a typo?

eguzki: fixed

jmprusi (approved):

PR #41, opened 2021-07-14
Basic rate limit

eguzki:
Infra basic rate limit (before authentication) for all endpoints exposed by the underlying APIs. All endpoints share the same bucket, so if rate limit exceeded for one endpoint, it will also exceed for the remaining endpoints of the same APIProduct.

```yaml
kind: APIProduct
metadata:
  name: animaltoys
spec:
  information:
    owner: whoever@mycompany.com
  routing:
    hosts:
    expose: true
  APIs:
    - name: dogs
      namespace: default
      tag: production
      mapping:
        prefix: /dogs
    - name: cats
      namespace: default
      tag: production
      mapping:
        prefix: /cats
  securityScheme:
    - name: MyAPIKey
      apiKeyAuth:
        name: APIKEY
        credential_source:
          labelSelectors:
            api: animaltoys
  rateLimit:
    max_value: 10
    period_in_seconds: 60
```

It reconcilliates changes in the `rateLimit` object, and even removing the `rateLimit` section will remove global rate limit.

It is specifically designed to be applied **before** auth http filter.


jmprusi:
What's the rate limit based on? client IP?

eguzki:
> What's the rate limit based on? client IP?

Currently *nothing*. Based on *hits* from anywhere, anyone and anyhow. And only one shared counter for all the operations (method + path) for all the APIs included in the APIProduct.

What would be "nice to have"?

jmprusi:
> > What's the rate limit based on? client IP?
> 
> Currently _nothing_. Based on _hits_ from anywhere, anyone and anyhow. And only one shared counter for all the operations (method + path) for all the APIs included in the APIProduct.
> 
> What would be "nice to have"?

Ok, right now we are basically allowing one single actor to DoS the endpoint by just sending enough traffic to block ALL client traffic to it. 

Can we define a rate limit based on some "source" information?  what are the options?

eguzki:
> e define a rate limit based on some "source" information? what are the options?

I thought it as a hard limit, true infra rate limit. We can additionally add limits based on source info, like the source address. 

If there is only limits based on source ip address, lets say X rps, then N IPs could create 2 * (N-1) rps and DoS on the upstream for saturation.

Yes, we can define a rate limit based on some "source" information. 
Rate limit logic is based on
* Requested "descriptor entries" created by Envoy based on [actions conf]
(https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-msg-config-route-v3-ratelimit-action)
* Definition of the limits based on those descriptors. 

What descriptors can Envoy generate from "source" information? 
* source_cluster
* remote_address
* any info in the HTTP request (path, host, method, ...)

PR #44, opened 2021-08-03
Added OIDC discovery feature

eloycoto:
When using annotated service, some annotations can make that get the OAS
    spec from the upstream API. When these annotations are in place, the
    controller will make an HTTP request to the defined endpoint and
    retrieve the config.

    Regarding testing, I went to unit Test instead of integration test. The
    main reason is that the controller needs to be inside the Kubernetes
    cluster to be able to use the internal service, so integration testing,
    as it is, is complicated.

    Unit testing is based on Golang Unit testing -no framework at all- soon;
    maybe we should think about moving this to Ginkgo or anything similar to
    this.


Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM golang:1.15 as builder
```
eguzki, line 2: This is OK, but `go.mod` also should be updated.

controllers/service_controller.go
```diff
@@ -110,6 +115,56 @@ func (r *ServiceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+			}
+		}
+
+		// This is terrible hack, we should use the same way as Coredns use it, but
+		// for the moment is ok like this
+		var targetURL = fmt.Sprintf("http://%s.%s.svc.cluster.local:%d", service.Name, service.Namespace, targetPort)
+		resp, err := http.Get(fmt.Sprintf("%s%s", targetURL, oasURLdefined))
+		if err != nil {
+			return true, "", err
+		}
```
eguzki, line 154: Missing `defer resp.Body.Close()`, otherwise there will be resource leaks. Connections will not be closed.

controllers/service_controller.go
```diff
@@ -110,6 +115,56 @@ func (r *ServiceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+		if resp.StatusCode != 200 {
+			return true, "", fmt.Errorf("cannot retrieve OAS from %s", targetURL)
```
eguzki, line 156: It would be nice to include in the error both the response code and the body content (which generally tries to explain error reason)

controllers/service_controller.go
```diff
@@ -110,6 +115,56 @@ func (r *ServiceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+					break
+				}
+			}
+		} else {
+			if len(service.Spec.Ports) > 0 {
+				targetPort = service.Spec.Ports[0].Port
+			}
+		}
+
+		// This is terrible hack, we should use the same way as Coredns use it, but
```
eguzki, line 148: how does CoreDNS use it? How could it be improved?

controllers/service_controller_unit_test.go
```diff
@@ -0,0 +1,253 @@
...
+		},
+	}
+}
+
+// TestOasDiscoveryNoAnnotation test that check if no annotation nothing
+// happens at all
+func TestOasDiscoveryNoAnnotation(t *testing.T) {
+	serviceReconciler := getServiceReconciler()
+	svc := getSampleService()
+	log := serviceReconciler.Logger().WithValues("test", "foo")
```
eguzki, line 122: minor thing, but I think it is better to use a different logger on testing code. something like:
```go
import (
    ctrl "sigs.k8s.io/controller-runtime" 
)
log = ctrl.Log.WithName("test")

File: controllers/service_controller.go [hunk 6]
```diff
@@ -110,6 +115,56 @@ func (r *ServiceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	return ctrl.Result{}, nil
 }
 
+func (r *ServiceReconciler) IsOasDefined(ctx context.Context, service *corev1.Service, log logr.Logger) (bool, string, error) {
```
eguzki, line 118: this method looks like it should be private. It is only used internally.

controllers/service_controller_unit_test.go
```diff
@@ -0,0 +1,253 @@
...
+	hasOas, result, err := serviceReconciler.IsOasDefined(context.Background(), svc, log)
```
eguzki, line 123: The testing approach is generally ok. It is testing main use cases, without annotations, with OAS in configmap and with OAS served by upstream. 

However it is being tested calling a `IsOasDefined` method that should, IMO, be private, hence non testeable. Another approach would be instantiate the reconciler and call `Reconcile()` method and then read current "fake cluster" status for expected results. We have been using this testing approach in other controllers:

```go
 endLoop := false                                                    
 for i := 0; i < 100 && !endLoop; i++ {                              
     res, err := r.Reconcile(req)                                    
     if err != nil {                                                 
         t.Fatal(err)                                                
     }                                                               
                                                                     
     endLoop = !res.Requeue                                          
 }                                                                   
                                                                     
 if !endLoop {                                                       
     t.Fatal("reconcile did not finish end of reconciliation as expected")
 }                                                                   
```

For the edge case where the endpoint is not available, the test can check the error returned by the reconcile and make sure the error matches. WDTY? it makes sense?

eguzki (changes requested):

Dockerfile
eguzki: an maybe in another PR

eguzki:
Add some doc showing the added service discovery annotations

Dockerfile
eloycoto: I do not think that deserves a new PR, the reason of the change is the io.ReadAll change, so I would keep in this PR, but I'm happy to make a new PR with a line change :D (It's already on a different commit)

controllers/service_controller_unit_test.go
eloycoto: The test of reconcile is used on some integration test, does not make much sense to me to have the unit test regarding that. I mean, OAS config function is quite simple, and has 100% coverage.

README.md
```diff
@@ -19,6 +19,8 @@ Kuadrant can leverage annotations. A good place to annotate is the Service used
 - **discovery.kuadrant.io/oas-configmap**: *OPTIONAL* Configmap that contains the OAS spec.
 - **discovery.kuadrant.io/matchpath**: *OPTIONAL* Define a single specific path, prefix or regex. Defaults to `/`.
 - **discovery.kuadrant.io/matchpath-type**: *OPTIONAL* Specifies how to match against the `matchpath` value. Accepted values are `Exact`, `Prefix` and `RegularExpression`. Defaults to `Prefix`.
+- **discovery.kuadrant.io/oas-path**: *OPTIONAL* Define a specific path for retrieving the config from the service itself.
+- **discovery.kuadrant.io/oas-name-port**: The port to be used to retrieve the OAS config, if not defined, it will used the first one
```
eguzki, line 23: missing `OPTIONAL` tag

README.md
```diff
@@ -90,6 +92,71 @@ spec:
       targetPort: 3000
 ```
 
+Example of a kuadrant annotated service providing OpenAPI spec in the service.
+
+```
```
eguzki, line 97:
```
```yaml 
```
for syntax highlight

controllers/service_controller_unit_test.go
eguzki: Up to you. But `IsOasDefined` is being public just to be tested in the unittests and looks like it should be private. That is a bad smell that something does not fit well.

controllers/service_controller_unit_test.go
eloycoto: Fixed.

controllers/service_controller.go
eloycoto: Fixed

controllers/service_controller.go
eloycoto: Fixed.

eloycoto:
please have a second look @eguzki

controllers/service_controller.go
```diff
@@ -110,6 +115,59 @@ func (r *ServiceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+		var targetURL = fmt.Sprintf("http://%s.%s.svc.cluster.local:%d", service.Name, service.Namespace, targetPort)
+		resp, err := http.Get(fmt.Sprintf("%s%s", targetURL, oasURLdefined))
+		if err != nil {
+			return true, "", err
+		}
+		if resp.StatusCode != 200 {
+			return true, "", fmt.Errorf("cannot retrieve OAS from '%s' statusCode='%d'", targetURL, resp.StatusCode)
+		}
+		body, err := io.ReadAll(resp.Body)
+		defer resp.Body.Close()
```
eguzki, line 161: It should be
```go
		resp, err := http.Get(fmt.Sprintf("%s%s", targetURL, oasURLdefined))
		if err != nil {
			return true, "", err
		}
                defer resp.Body.Close()
```

https://medium.com/@KeithAlpichi/go-gotcha-closing-a-nil-http-response-body-with-defer-9b7a3eb30e8c

controllers/service_controller.go
eguzki: Sorry, I do not fully get it. We really need to inspect the service type and if
* service type is external name: get the CNAME
* service is headless: get the endpoints
? If that is the case, the controller-runtime should have this already implemented. Maybe we can use it?

I just want to understand what it the best way to do this.

eguzki (changes requested):

controllers/service_controller_unit_test.go
eguzki: I do not see the fix

eguzki: I do now

eguzki:
looking good, I left some comments

Dockerfile
eguzki: It should be changed in the go.mod as well

controllers/service_controller.go
eloycoto: I thought that we agreed on private channel on this hack, I'm not sure if you want me to implement the right way, with const, externalName, etc.. 

Regards

eloycoto:
Updated the defer, ready for review.

eguzki:
still some comments unanswered

controllers/service_controller.go
```diff
@@ -110,6 +117,69 @@ func (r *ServiceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+		// Context: https://github.com/Kuadrant/kuadrant-controller/pull/44#discussion_r684146743
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return true, "", fmt.Errorf("cannot read the body of %s: %w", targetURL, err)
+		}
+
+		if resp.StatusCode != 200 {
+			var bodyLogMax = maxBodyLog
+			if cap(body) < bodyLogMax {
+				bodyLogMax = cap(body)
```
eguzki, line 172: use `len`. `cap` is the capacity that the array/slice has currently to host data. `len` and `cap` are not necessarily equal.

eloycoto: The thing is, you don't want that bodyLogMax is bigger than capacity, so that it triggers a panic. This is why I use cap instead of len, I'll change

eloycoto:
To avoid the drama with the CoreDNS way, I added a new commit, that fix the hack. 

1) It uses name, and namespace, it avoids cluster name, because is part the resolv.conf. Will not work with k8s federation.
2) If externalName is used, it will use it. 

There a few things are missed, that I'm not sure if you want to track now: 

- IPv4 fallback when IPv6 fails, maybe the dual-stack is not working as expected, and only IPv4 need to be used. 
- Headless services: multiple IPS can be returned on a SRV record, request should go to the next service if the first fails. 
- And most important one, if SSL is present (On TCP level), we should fallback and use https, looks like that you don't really care here, but I think that this is a use case. 

Hope these latest commits fulfil your expectations on this feature.

Best regards

eguzki:
> To avoid the drama with the CoreDNS way, I added a new commit, that fix the hack.
> 
>     1. It uses name, and namespace, it avoids cluster name, because is part the resolv.conf. Will not work with k8s federation.
I think this is the proper way to access a service by domain name. By name and namespace. Just that. The `cluster.local` is the default cluster domain assigned, but it could be otherwise like: `cluster-domain.example` or whatever the kubernetes cluster has been configured. So, `<service-name>.<namespace>` is more than enough and does not depend on the cluster domain. `<service-name>.<namespace>.svc` is also valid, as cluster domain is set in the `/etc/resolv.conf` by kubelet as search domain.

More info here: https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/

> 
>     2. If externalName is used, it will use it.
No need to do this. That code can be removed. When services with external name are created, kubernetes adds a CNAME entry and the name is resolved correctly. I have tried myself with this service and works like a charm in local cluster:
```yaml
apiVersion: v1                      
kind: Service                       
metadata:                           
  name: my-echo-api                 
  namespace: ns-a                   
spec:                               
  type: ExternalName                
  externalName: echo-api.3scale.net 
```

> 
> 
> There a few things are missed, that I'm not sure if you want to track now:
> 
>     * IPv4 fallback when IPv6 fails, maybe the dual-stack is not working as expected, and only IPv4 need to be used.
I think this is out of scope
> 
>     * Headless services: multiple IPS can be returned on a SRV record, request should go to the next service if the first fails.

When headless services are created, both A/AAAA and SRV DNS records are created. SRV records are used to resolve DNS names with named ports, otherwise A record is used. The proposed solution works out of the box for "normal", headless and external name services

More info here: https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/

> 
>     * And most important one, if SSL is present (On TCP level), we should fallback and use https, looks like that you don't really care here, but I think that this is a use case.
> 

Good point. Maybe add a new optional annotation
```
discovery.kuadrant.io/oas-scheme
```
which is HTTP by default, but allows to opt-in to HTTPS.

> 
> Hope these latest commits fulfil your expectations on this feature.
> 
> Best regards

eloycoto:
IMHO, This is getting ridiculous, that for something so simple we're doing a
lot of overengineering:

> 
> 
>     2. If externalName is used, it will use it.
> 
> No need to do this. That code can be removed. When services with external name are created, kubernetes adds a CNAME entry and the name is resolved correctly. I have tried myself with this service and works like a charm in local cluster:
> 
> apiVersion: v1                      
> kind: Service                       
> metadata:                           
>   name: my-echo-api                 
>   namespace: ns-a                   
> spec:                               
>   type: ExternalName                
>   externalName: echo-api.3scale.net 
> 
> / # dig +search my-echo-api
> 
> ; <<>> DiG 9.11.6-P1 <<>> +search my-echo-api
> ;; global options: +cmd
> ;; Got answer:
> ;; WARNING: .local is reserved for Multicast DNS
> ;; You are currently testing what happens when an mDNS query is leaked to DNS
> ;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 46142
> ;; flags: qr aa rd; QUERY: 1, ANSWER: 5, AUTHORITY: 0, ADDITIONAL: 1
> ;; WARNING: recursion requested but not available
> 
> ;; OPT PSEUDOSECTION:
> ; EDNS: version: 0, flags:; udp: 4096
> ; COOKIE: 1e853ef722f45eb0 (echoed)
> ;; QUESTION SECTION:
> ;my-echo-api.ns-a.svc.cluster.local. IN	A
> 
> ;; ANSWER SECTION:
> my-echo-api.ns-a.svc.cluster.local. 30 IN CNAME	echo-api.3scale.net.
> echo-api.3scale.net.	30	IN	CNAME	echo-api.pro-saas.3sca.net.
> echo-api.pro-saas.3sca.net. 30	IN	A	54.157.201.86
> echo-api.pro-saas.3sca.net. 30	IN	A	3.213.166.152
> echo-api.pro-saas.3sca.net. 30	IN	A	52.72.133.221
> 
> ;; Query time: 55 msec
> ;; SERVER: 10.96.0.10#53(10.96.0.10)
> ;; WHEN: Tue Aug 10 12:22:15 UTC 2021
> ;; MSG SIZE  rcvd: 327

I'm ok with this sentence, but the CNAME will resolve to this A IP, that's ok..
but the host header will be `name.nemspace.svc.cluster.local` when the user
defined a externalService, if we use the internal one, the Host header of the
HTTP request will the internal one, with that, some failures can happen:

1) Host header check on the Upstream service does not match, so things got rejected.
2) TLS, if defined https, can fail on the SNI check. 

And, I'm saying this, because I suffered this situation on APICast.

>When headless services are created, both A/AAAA and SRV DNS records are created. SRV records are used to resolve DNS names with named ports, otherwise A record is used. The proposed solution works out of the box for "normal", headless and external name services
>
>More info here: https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/
>

Ok, should we retry, no? should we keep track of the dns record, because that
maybe is important. Up to you

> Good point. Maybe add a new optional annotation
> 
> discovery.kuadrant.io/oas-scheme
> 
> which is HTTP by default, but allows to opt-in to HTTPS.

Ok, do you want to have in this PR, or do you want to have in a separate one?
If you need in this PR, should we add another annotations for this:

`discovery.kuadrant.io/oas-tls-verify`: to enable/verify the certificate verification.
`discovery.kuadrant.io/oas-ca-certificate-configmap`: to get the CA certificate
from a CA, because maybe is a private CA.

Are you ok with all other changes? It's not clear to me. 
Regards

eguzki:
> but the host header will be name.nemspace.svc.cluster.local when the user
defined a externalService, if we use the internal one, the Host header of the
HTTP request will the internal one, with that, some failures can happen:

That is ok and expected. Otherwise, do not create externalName services.

> Ok, should we retry, no? should we keep track of the dns record, because that
maybe is important. Up to you

Headless services are working out of the box, there is nothing we need to implement in this regard.

> Ok, do you want to have in this PR, or do you want to have in a separate one?
If you need in this PR, should we add another annotations for this:
> discovery.kuadrant.io/oas-tls-verify: to enable/verify the certificate verification.
> discovery.kuadrant.io/oas-ca-certificate-configmap: to get the CA certificate
from a CA, because maybe is a private CA.

It can be added later. Up to you. For the implementation, I would say that ` discovery.kuadrant.io/oas-tls-verify` mechanism is very useful for devtesting, but not recommendable for prod envs allowing the customer to disable tls verification. For now, I would configure the HTTPS client to not verify TLS connection with a big `TODO` to change it in the future. We have already done that several times.

eloycoto:
Then to clarify, I'm a bit lost right now:

1) Just remove the getServiceDomainName function, and always use the name.namespace.svc. Added the caveat that will not work with ExternalNames. 
2) Added the TLS option, and insecure by default, no? Documentation here.
3) Regarding other comments, are all as you want? Or want more changes? I think that I addressed all, but this PR is now >40 comments, and maybe I miss something. 

> Headless services are working out of the box, there is nothing we need to implement in this regard.

Well, I'm not fully agree on this, but I'm not the reviewer here.

eguzki (approved):

PR #46, opened 2021-08-17
Pre-auth rate limit

eguzki:
Pre-auth rate limit for all endpoints exposed by the underlying APIs. All endpoints share the same bucket, so if rate limit exceeded for one endpoint, it will also exceed for the remaining endpoints of the same APIProduct.

```yaml
kind: APIProduct
metadata:
  name: animaltoys
spec:
  information:
    owner: whoever@mycompany.com
  routing:
    hosts:
    expose: true
  APIs:
    - name: dogs
      namespace: default
      tag: production
      mapping:
        prefix: /dogs
    - name: cats
      namespace: default
      tag: production
      mapping:
        prefix: /cats
  securityScheme:
    - name: MyAPIKey
      apiKeyAuth:
        name: APIKEY
        credential_source:
          labelSelectors:
            api: animaltoys
  rateLimit:
    max_value: 10
    period_in_seconds: 60
```

Implementations details:
* It reconcilliates changes in the `rateLimit` object, and even removing the `rateLimit` section will remove global rate limit.
* It is specifically designed to be applied **before** auth http filter.
* Rate limit is based on `remote address`
* [Global rate limit](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/other_features/global_rate_limiting)
* [Limitador](https://github.com/3scale-labs/limitador) as the gRPC rate limiting service
* [Limitador Operator](https://github.com/3scale-labs/limitador-operator/) used to deploy Limitador and manage rate limit configurations.


config/rbac/role.yaml
```diff
@@ -35,6 +35,18 @@ rules:
...
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
 - apiGroups:
   - networking.istio.io
   resources:
```
jmprusi, line 52: Missing `envoyfilters` perms

eguzki: Good catch!

eguzki: fixed!

andrewdavidmackenzie:
OK, so when I read "infra rate limit" I understood this would protect the service/infra by offering one global limit that cannot be exceeded. However, I read mention of it being per "Remote IP" and hence not being a limit you can rely on ... but the total traffic will be a function of the number of unique remote IPs used....

Can you use something like "*" to group all remote IPs into a single bucket and hence get the same affect: knowing that under no circumstances will the requests to the protected service exceed "X" rps?

eguzki:
> OK, so when I read "infra rate limit" I understood this would protect the service/infra by offering one global limit that cannot be exceeded. However, I read mention of it being per "Remote IP" and hence not being a limit you can rely on ... but the total traffic will be a function of the number of unique remote IPs used....
> 
> Can you use something like "*" to group all remote IPs into a single bucket and hence get the same affect: knowing that under no circumstances will the requests to the protected service exceed "X" rps?

This what I have called "infra hard rate limit". From the [analysis doc](https://docs.google.com/document/d/1oLZeXLf30WlOD1TNFhrVJoZdgsZOMdtIXw4GnRbYQkk/edit).

This "infra hard rate limit" was implemented in this *not merged* PR #41. That is why we call the rate limiit implemented in this PR (per remote IP) as *soft limit*. Anyone of both approaches can be implemented, or even both at the same time, which also makes sense to me, but maybe not to others. That is what we need to know.. what is really wanted?

andrewdavidmackenzie:
OK.

IMHO, once you move away from *one* limit, that is global and really protects the infrastructure.... I think they are no longer "infra" limits, and maybe best to look for a new name.

For these type of limits, it would be interesting to make as flexible as possible (eventually...) and to be based on anythign available to the enforcement point (prior to auth and user/app/api identification) such as headers, etc and not just a single concept (like remote IP)?

eguzki:
This is the interesting feedback requested in the [analysis doc](https://docs.google.com/document/d/1oLZeXLf30WlOD1TNFhrVJoZdgsZOMdtIXw4GnRbYQkk/edit#heading=h.x62kxierd9q)

> IMHO, once you move away from one limit, that is global and really protects the infrastructure.... I think they are no longer "infra" limits, and maybe best to look for a new name.

I disagree. Rate limiting based on IP is also a kind of rate limit to protect the infrastructure. Authenticated rate limiting is, on the other hand, used to limit usage on clients for business purposes. 

But I am ok to look for a new name. 

> For these type of limits, it would be interesting to make as flexible as possible (eventually...) and to be based on anythign available to the enforcement point (prior to auth and user/app/api identification) such as headers, etc and not just a single concept (like remote IP)?

:+1: For the first milestone, IMO a hard limit and/or a soft limit (based on remote IP) is sufficient.

apis/networking/v1beta1/apiproduct_types.go
```diff
@@ -87,6 +89,11 @@ func (a *APISelector) APINamespacedName() types.NamespacedName {
 	return types.NamespacedName{Namespace: a.Namespace, Name: name}
 }
 
+type InfraRateLimitSpec struct {
+	MaxValue int32 `json:"max_value"`
```
guicassolato, line 93:
```suggestion
	MaxValue int32 `json:"maxValue"`
```

apis/networking/v1beta1/apiproduct_types.go
```diff
@@ -87,6 +89,11 @@ func (a *APISelector) APINamespacedName() types.NamespacedName {
...
+	Period   int32 `json:"period_in_seconds"`
```
guicassolato, line 94:
```suggestion
	Period   int32 `json:"period"`
```

I'd call it just "period". We can specify that it should be set "in seconds" in the description of the field. Later in the future, if we want we can support humanized values as well, e.g. "1s", "3d", "1m", etc.

apis/networking/v1beta1/apiproduct_types.go
eguzki: :+1:  I will update the manifests as well

eguzki:
requested changes applied

eguzki:
closing in favor  of #47  

The contents of this PR will be in #47

PR #47, opened 2021-08-18
Authenticated rate limit

eguzki:
Authenticated basic rate limit (after authentication) for all endpoints exposed by the underlying APIs. All endpoints share the same bucket, so if rate limit exceeded for one endpoint, it will also exceed for the remaining endpoints of the same APIProduct.

```yaml
kind: APIProduct
metadata:
  name: animaltoys
spec:
  information:
    owner: whoever@mycompany.com
  routing:
    hosts:
    expose: true
  APIs:
    - name: dogs
      namespace: default
      tag: production
      mapping:
        prefix: /dogs
    - name: cats
      namespace: default
      tag: production
      mapping:
        prefix: /cats
  securityScheme:
    - name: MyAPIKey
      apiKeyAuth:
        name: APIKEY
        credential_source:
          labelSelectors:
            api: animaltoys
  rateLimit:
    authenticated:
      max_value: 5
      period_in_seconds: 30
    preAuthenticated:
      max_value: 10
      period_in_seconds: 60
```

Implementations details:
* It reconcilliates changes in the `rateLimit.authenticated` object. Removing the `rateLimit.authenticated` section, the controller will disable auth rate limit.
* It is specifically designed to be applied **after** auth http filter.
* Rate limit is based on `user_id`
* Integration with APIKey and OIDC authN
* [Global rate limit](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/other_features/global_rate_limiting)
* [Limitador](https://github.com/3scale-labs/limitador) as the gRPC rate limiting service
* Upgraded Authorino to v0.4.0

Depends on #46


apis/networking/v1beta1/apiproduct_types.go
```diff
@@ -89,11 +89,26 @@ func (a *APISelector) APINamespacedName() types.NamespacedName {
...
 
-type InfraRateLimitSpec struct {
+type AuthRateLimitSpec struct {
 	MaxValue int32 `json:"max_value"`
 	Period   int32 `json:"period_in_seconds"`
 }
 
+type PreAuthRateLimitSpec struct {
+	MaxValue int32 `json:"max_value"`
+	Period   int32 `json:"period_in_seconds"`
```
guicassolato, line 99: I've left a couple comments about these in #46.

apis/networking/v1beta1/apiproduct_types.go
```diff
@@ -89,11 +89,26 @@ func (a *APISelector) APINamespacedName() types.NamespacedName {
...
+}
+
+type RateLimitSpec struct {
+	// PreAuthRateLimitSpec configures rate limit parameters per remote address
+	// +optional
+	PreAuthRateLimit *PreAuthRateLimitSpec `json:"preAuthenticated,omitempty"`
```
guicassolato, line 105: I think we should call it something else. "PreAuthRateLimit" makes sense if know that it opposes to "AuthenticatedRateLimit". If the user is planning on using "auth", speaking of "pre-auth" sounds weird IMO.

Maybe we should call this one just "RateLimit"?

I also like the idea (once suggested by @jmprusi) of defining "phases".

apis/networking/v1beta1/apiproduct_types.go
guicassolato: Can we reuse the struct already defined?

eguzki: Sure, I will propagate the changes

eguzki:
Closing in favor of #48

apis/networking/v1beta1/apiproduct_types.go
eguzki: > Can we reuse the struct already defined?

we can. I deliberately duplicated the struct to make explicit they are different things and can have different structure and fields. If the design gets stabilized, I will remove duplication (if it exists)

PR #56, opened 2021-10-07
make security requirements optional

eguzki:
Support for APIProduct obejcts without security (AuthN) configuration

Reconcile adding and removing security from apiproduct

This PR is based on #55, it is ready for review, but should be merged after it

### DIFF
--- a/pkg/authproviders/authorino/provider.go
+++ b/pkg/authproviders/authorino/provider.go
@@ -158,17 +158,19 @@ func (a *Provider) Status(ctx context.Context, apip *networkingv1beta1.APIProduc
 	// If any object is missing/not-created, Status returns false.
 	authConfig := buildAuthConfig(apip)
 
-	existingAuthConfig := &authorino.AuthConfig{}
-	err := a.Client().Get(ctx, client.ObjectKeyFromObject(authConfig), existingAuthConfig)
-	if err != nil && errors.IsNotFound(err) {
-		return false, nil
-	} else if err != nil {
-		return false, err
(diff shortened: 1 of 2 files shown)


apis/networking/v1beta1/apiproduct_types.go
```diff
@@ -106,8 +106,11 @@ type APIProductSpec struct {
 	// platform, short-names can also be used instead of a FQDN (i.e. has no
 	// dots in the name). In such a scenario, the FQDN of the host would be
 	// derived based on the underlying platform.
-	Hosts          []string          `json:"hosts"`
-	SecurityScheme []*SecurityScheme `json:"securityScheme"`
+	Hosts []string `json:"hosts"`
+
+	// Configure authentication mechanisms
+	// +optional
```
guicassolato, line 112: Is this needed for controller-gen to make it optional? I usually rely only on `omitempty` and it does the trick, but I'd love to learn about any other extra advantage.

guicassolato (approved):
I've left a couple comments kicking off conversation aside, but nothing that disapproves the change of course. LGTM

apis/networking/v1beta1/apiproduct_types.go
eguzki: I have experienced that too. That's confusing as the [kubebulder doc](https://book.kubebuilder.io/reference/markers/crd-validation.html) says it should be the `'+optional` marker.

To make it explicit and easy to read, I always add it for optional fields

eguzki:
Added fix for the apiproduct status subresource computing logic when resources are optional.

PR #57, opened 2021-10-08
add event create permission to the manager's role

eguzki:
The controllers create namespaced events when there are unexpected errors. 

This PR adds the required permissions to the role assigned to the controller when deployed.


controllers/apiproduct_controller.go
```diff
@@ -153,6 +154,7 @@ func (r *APIProductReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return statusResult, nil
 	}
 
+	r.EventRecorder().Event(apip, corev1.EventTypeWarning, "this is a reason", "event message")
```
guicassolato, line 157: From https://pkg.go.dev/k8s.io/client-go@v0.21.2/tools/record#EventRecorder

> [...] 'reason' should be short and unique; it should be in UpperCamelCase format (starting with a capital letter). "reason" will be used to automate handling of events, so imagine people writing switch statements to handle them. You want to make that easy.

🙂

PR #67, opened 2021-11-18
Add a named port

maleck13:
istio needs ports to indicate which type of protocol they use. Adding this named port allows for the WASMPlugin to work with the sidecar


eguzki (approved):

eguzki:
Feel free to squash and merge

PR #69, opened 2021-12-03
WIP Wasm plugin support

maleck13:
Adds new providers and a walkthough. THis is only meant as a POC currently and needs futher work before merging


maleck13:
Will look into some of these failures

maleck13:
Ok one of the annotations we use on the service changed in gateway api 0.4.0 which is what was causing the issue. This is a non backwards compatible change

utils/local-deployment/local-setup.sh
```diff
@@ -20,17 +20,29 @@ set -euo pipefail
...
 
+echo "setting up kuadrant for provider ${PROVIDER}"
+
 export KUADRANT_NAMESPACE="kuadrant-system"
 export KIND_CLUSTER_NAME="kuadrant-local"
 
 ${SCRIPT_DIR}/deploy-kuadrant-deps.sh
 
-echo "Building kuadrant"
+echo "Building kuadrant ${PROVIDER}"
```
eguzki, line 30: this should go in `deploy-kuadrant-deps.sh`

The reason is that local-setup.sh is mean for deploying the controller as a pod. 

But for dev testing, sometimes it is handy to run it locally (outside the cluster), and the kuadrant infra (istio, limitador, authorino) is still needed to be there, so `deploy-kuadrant-deps.sh` is being reused for both deployment modes

utils/local-deployment/local-setup.sh
```diff
@@ -20,17 +20,29 @@ set -euo pipefail
...
+  kubectl apply -f utils/local-deployment/istio-manifests/Base/gateway-api/gateway.yaml
+fi
+
+
 docker build -t kuadrant:devel ./
 kind load docker-image kuadrant:devel --name ${KIND_CLUSTER_NAME}
 echo "Deploying Kuadrant control plane"
 kustomize build config/default | kubectl -n "${KUADRANT_NAMESPACE}" apply -f -
-kubectl -n "${KUADRANT_NAMESPACE}" patch deployment kuadrant-controller-manager -p '{"spec": {"template": {"spec":{"containers":[{"name": "manager","image":"kuadrant:devel", "env": [{"name": "LOG_LEVEL", "value": "debug"}, {"name": "LOG_MODE", "value": "development"}], "imagePullPolicy":"IfNotPresent"}]}}}}'
+kubectl -n "${KUADRANT_NAMESPACE}" patch deployment kuadrant-controller-manager -p '{"spec": {"template": {"spec":{"containers":[{"name": "manager","image":"kuadrant:devel", "env": [{"name": "LOG_LEVEL", "value": "debug"}, {"name": "LOG_MODE", "value": "development"},{"name": "KUADRANT_INGRESS_PROVIDER", "value": "'$PROVIDER'" }], "imagePullPolicy":"IfNotPresent"}]}}}}'
```
eguzki, line 45: For running locally, the makefile target `run` should also export `KUADRANT_INGRESS_PROVIDER` env var

pkg/authproviders/authorino/wasm.go
```diff
@@ -0,0 +1,137 @@
...
+				},
+			},
+			// The hard coded pieces here are purly for POC reasons
+			Spec: v1alpha1.WasmPlugin{
+				Selector:   selector,
+				Url:        "https://raw.githubusercontent.com/Kuadrant/wasm-shim/main/deploy/wasm_shim.wasm",
+				PluginName: "apim-shim",
+				PluginConfig: &types.Struct{
+					Fields: map[string]*types.Value{
+						"auth_cluster":      {Kind: &types.Value_StringValue{StringValue: "outbound|50051||authorino-authorization.kuadrant-system.svc.cluster.local"}},
```
eguzki, line 72: tiny thing, `cluster.local` suffix might not be true for all clusters

pkg/authproviders/authorino/wasm.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	if !reflect.DeepEqual(desired.Spec, existing.Spec) {
+		existing.Spec = desired.Spec
+		updated = true
+	}
+
+	return updated, nil
+}
+
+func (wp *WasmProvider) buildSelector(ctx context.Context, api *networkingv1beta1.API) (*v1beta1.WorkloadSelector, error) {
```
eguzki, line 105: Some food for though: If the service is of type `ExternalName`, there are no pods backing the service. ExternalName typed services were the way to go if the APIManagement is done in some cluster A, but the service is external (URL).

For that specific case, this wasm plugin would not work, would it?

PR #70, opened 2021-12-15
SDK updates and bundle generation

mikenairn:
**Updates for operator SDK version v1.15.0**
Update dependencies https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.14.0
Update Makefile targets to be consistent with latest version.

Set IMAGE_TAG_BASE to quay.io/kuadrant/kuadrant-controller
Removed custom envtest setup, use ENVTEST_K8S_VERSION 1.22

**Fix issue with bundle generation** 
The kustomize command fails, but the bundle make target still continues resulting in a bundle being created with only a CSV and no additional files.

```
$ ./bin/kustomize build config/manifests
Error: accumulating resources: 2 errors occurred:
        * accumulateFile error: "accumulating resources from 'bases/networking.kuadrant.io_apis.yaml': evalsymlink failure on '/home/mnairn/go/src/github.com/kuadrant/kuadrant-controller/config/manifests/bases/networking.kuadrant.io_apis.yaml' : lstat /home/mnairn/go/src/github.com/kuadrant/kuadrant-controller/config/manifests/bases/networking.kuadrant.io_apis.yaml: no such file or directory"
        * loader.New error: "error loading bases/networking.kuadrant.io_apis.yaml with git: url lacks host: bases/networking.kuadrant.io_apis.yaml, dir: evalsymlink failure on '/home/mnairn/go/src/github.com/kuadrant/kuadrant-controller/config/manifests/bases/networking.kuadrant.io_apis.yaml' : lstat /home/mnairn/go/src/github.com/kuadrant/kuadrant-controller/config/manifests/bases/networking.kuadrant.io_apis.yaml: no such file or directory, get: invalid source string: bases/networking.kuadrant.io_apis.yaml"
```
**Generate bundle**
```
make bundle
```


eguzki:
Overall looks good. The bundle generation was broken from the very beginning, but it was unused so we did not care.

One question about this. Kuadrant controller is not going to be released independently as OLM bundle inside a catalog. What is the use of having CSV manifest? I think that the manifest bundle of the kuadrant controller should consist of:
* Deployment object
* Roles, ClusterRoles, RoleBinding, ClusterRoleBindings, SA (maybe)
* CRDs
* ServiceMonitor (metrics)
* Service

Makefile
```diff
@@ -140,22 +160,70 @@ rm -rf $$TMP_DIR ;\
 }
 endef
 
-# Generate bundle manifests and metadata, then validate generated files.
 .PHONY: bundle
-bundle: manifests kustomize
+bundle: manifests kustomize ## Generate bundle manifests and metadata, then validate generated files.
 	operator-sdk generate kustomize manifests -q
```
eguzki, line 165: The operator-sdk should be pinned to a version. There should be a target to install the tool in `./bin` and used from there.

Makefile
```diff
@@ -140,22 +160,70 @@ rm -rf $$TMP_DIR ;\
...
+##@ Misc
+
+## Miscellaneous Custom targets
+
+KUADRANT_NAMESPACE=kuadrant-system
+
 KIND = $(shell pwd)/bin/kind
-kind:
+kind: ## Download kind locally if necessary.
 	$(call go-get-tool,$(KIND),sigs.k8s.io/kind@v0.11.1)
```
eguzki, line 227: Upgrading operator-sdk (and controller-runtime and controller-gen) to 1.22, should we also upgrade kind to k8s 1.22?

Makefile
mikenairn: It's using v1.15.0, i agree though we should just put the version we use in the bin directory so it's pinned. This is just how the bog standard generated Makefile looks. Can look into pinning the sdk version separately.

mikenairn:
> Overall looks good. The bundle generation was broken from the very beginning, but it was unused so we did not care.
> 
> One question about this. Kuadrant controller is not going to be released independently as OLM bundle inside a catalog. What is the use of having CSV manifest? I think that the manifest bundle of the kuadrant controller should consist of:
> 
>     * Deployment object
> 
>     * Roles, ClusterRoles, RoleBinding, ClusterRoleBindings, SA (maybe)
> 
>     * CRDs
> 
>     * ServiceMonitor (metrics)
> 
>     * Service

@eguzki You are right, the controller will not be deployed independently, and won't need a catalog(index) image. The reason for wanting to have a bundle however is so that we can just include it in the kuadrant-operator bundle(it will have a dependency on OLM) and add the kuadrant-controller as a dependency of the kuadrant-operator.

Basically, OLM deals with installing/upgrading the kuadrant-controller k8s resources if we do it this way, the kuadrant operator just needs to reconcile the appropriate CRs for each of it's dependants based on it's configuration(kuadrant CR).

```
mnairn@deacon kuadrant-operator (main) $ BUNDLE_IMGS=quay.io/kuadrant/kuadrant-operator-bundle:v0.0.1,quay.io/kuadrant/kuadrant-controller-bundle:v0.0.1 make catalog-build
```

```
mnairn@deacon kuadrant-operator (main) $ cat bundle/metadata/dependencies.yaml 
dependencies:
  - type: olm.package
    value:
      packageName: authorino-operator
      version: "0.0.1"
  - type: olm.package
    value:
      packageName: limitador-operator
      version: "0.0.1"
  - type: olm.package
    value:
      packageName: kuadrant-controller
      version: "0.0.1"
```

eguzki:
> > Overall looks good. The bundle generation was broken from the very beginning, but it was unused so we did not care.
> > One question about this. Kuadrant controller is not going to be released independently as OLM bundle inside a catalog. What is the use of having CSV manifest? I think that the manifest bundle of the kuadrant controller should consist of:
> > ```
> > * Deployment object
> > 
> > * Roles, ClusterRoles, RoleBinding, ClusterRoleBindings, SA (maybe)
> > 
> > * CRDs
> > 
> > * ServiceMonitor (metrics)
> > 
> > * Service
> > ```
> 
> @eguzki You are right, the controller will not be deployed independently, and won't need a catalog(index) image. The reason for wanting to have a bundle however is so that we can just include it in the kuadrant-operator index image(it will have a dependency on OLM) and add the kuadrant-controller as a dependency of the kuadrant-operator.
> 
> Basically, OLM deals with installing/upgrading the kuadrant-controller k8s resources if we do it this way, the kuadrant operator just needs to reconcile the appropriate CRs for each of it's dependants based on it's configuration(kuadrant CR).
> 
> ```
> mnairn@deacon kuadrant-operator (main) $ BUNDLE_IMGS=quay.io/kuadrant/kuadrant-operator-bundle:v0.0.1,quay.io/kuadrant/kuadrant-controller-bundle:v0.0.1 make catalog-build
> ```
> 
> ```
> mnairn@deacon kuadrant-operator (main) $ cat bundle/metadata/dependencies.yaml 
> dependencies:
>   - type: olm.package
>     value:
>       packageName: authorino-operator
>       version: "0.0.1"
>   - type: olm.package
>     value:
>       packageName: limitador-operator
>       version: "0.0.1"
>   - type: olm.package
>     value:
>       packageName: kuadrant-controller
>       version: "0.0.1"
> ```

That would make kuadrant-controller a *de facto* a operator from the OLM point of view. It will be available in the catalog as a standalone operator and could be installed standalone. I wonder if that is wanted. I understand how easy is to install something through dependencies, saving the kuadrant operator to bear with the manifests. However, this way, you do not have control on the deployment of the kuadrant-controller. The OLM takes care of it. What if the operator wants to add env var to the kuadrant-controller? Maybe this topic deserves offline discussion.

Makefile
mikenairn: We've pinned the version of the oprator-sdk binary in other repos, but since we are no longer generating a bundle for the controller, and it's the only make target that uses it, it's probably not needed here.

mikenairn:
@eguzki Would you mind taking a look at this again. The bundle has now been removed in favour of generating a set of deploy manifests in the same way we are doing in other repos `config/deploy/manifests.yaml`. These will be pulled in by the operator.

eguzki (approved):

PR #74, opened 2022-01-07
Add finalizer

maleck13:


rahulanand16nov (approved):
LTGM.

Let's avoid the failing test for now since we are still changing the resource structure and make sure about them once we are ready to merge into main.

PR #76, opened 2022-01-20
tie networking and ratelimit using annotations

rahulanand16nov:
This PR:
- Move networking and ratelimit mapping out of API to annotations in RateLimitPolicy.
- Fixes `BOTH` stage value not turning into two ratelimit fields
- Makes uniqueness requirement on the route names in RateLimitPolicy.
- Adds annotation to the virtualservice when RateLimitPolicy is created to map back to it.
- Removes selectors from API resource.

## Steps to verify functionality
1. `kind create cluster`
2. `kuadrantctl install`
3. `kubectl apply -f examples/toystore/toystore.yaml`
4. `kubectl apply -f examples/toystore/virtualservice.yaml`
5. `make install`
6. `make run`
7. `kubectl apply -f config/samples/apim_v1alpha1_ratelimitpolicy.yaml`
8. `kubectl port-forward -n kuadrant-system svc/kuadrant-gateway 9080:80`
9. `curl -X POST -H "Host: toy.toystore.com" localhost:9080/admin/addToy -v` 2 times


config/samples/apim_v1alpha1_ratelimitpolicy.yaml
```diff
@@ -5,6 +5,8 @@ metadata:
   namespace: kuadrant-system
   labels:
     app: toystore
+  annotations:
```
maleck13, line 8: I think this should be part of the spec as per our discussion today 

```
targetRef:
   type: VirtualService
   name: myservice
```

rahulanand16nov: Updated. Just instead of `targetRef`, I have chosen `networkingRef`

controllers/apim/ratelimitpolicy_controller.go
```diff
@@ -85,15 +101,338 @@ func (r *RateLimitPolicyReconciler) Reconcile(eventCtx context.Context, req ctrl
...
-	logger.Info("spec reconcile done", "result", specResult, "error", specErr)
-	if specErr == nil && specResult.Requeue {
-		logger.Info("Reconciling not finished. Requeueing.")
-		return specResult, nil
+	if err := r.reconcileLimits(ctx, &rlp); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	vsName, ok := rlp.Annotations[RLPAnnotationVSName]
+	if !ok {
```
maleck13: Does this need to be removed?

rahulanand16nov: I guess that's the old commit, I cannot find that in the new commit.

rahulanand16nov: Okay, that constant was still present, removed.

controllers/apim/ratelimitpolicy_controller.go
```diff
@@ -85,15 +100,343 @@ func (r *RateLimitPolicyReconciler) Reconcile(eventCtx context.Context, req ctrl
...
+
+	for _, gw := range vs.Spec.Gateways {
+		gwKey := gatewayNameToObjectKey(gw, vs.Namespace) // Istio defaults to vs namespace
+		rateLimitPatch := rateLimitInitialPatch(gwKey)
+
+		if err := controllerutil.SetOwnerReference(rlp, rateLimitPatch, r.Client().Scheme()); err != nil {
+			logger.Error(err, "failed to add owner ref to RateLimit filters patch")
+			return err
+		}
+		if err := r.Client().Create(ctx, rateLimitPatch); err != nil {
```
eguzki, line 206: Objects needs to be reconciled, not only created. Otherwise changes in the RLP will not update the envoyfilter objects.

Anyway, it is good enough for now

rahulanand16nov: Yes, just waiting for the green light before i tighten those loose screws.

Reconciling EnvoyFilter is a bit tricky since need to be aware of VS/HTTPR change as well.

controllers/apim/ratelimitpolicy_controller.go
```diff
@@ -85,15 +100,343 @@ func (r *RateLimitPolicyReconciler) Reconcile(eventCtx context.Context, req ctrl
...
+			},
+		},
+	}
+
+	patchRaw, _ := json.Marshal(patchUnstructured)
+	stringPatch := string(patchRaw)
+
+	jsonActions, _ := json.Marshal(actions)
+	// A nice trick to make it easier add-in actions
+	stringPatch = strings.Replace(stringPatch, "\"ACTIONS\"", string(jsonActions), 1)
```
eguzki, line 373: This should not be done. Structured data should not be serialized to apply string substitution and then de-serialize 
again. 

You can build the unstructured (based on maps and interfaces) actions list object and then apply to the `patchUnstructured` object.


For now it is ok

controllers/apim/ratelimitpolicy_controller.go
```diff
@@ -85,15 +100,343 @@ func (r *RateLimitPolicyReconciler) Reconcile(eventCtx context.Context, req ctrl
...
 }
 
+func (r *RateLimitPolicyReconciler) reconcileWithVirtualService(ctx context.Context, vs *istio.VirtualService, rlp *apimv1alpha1.RateLimitPolicy) error {
+	logger := r.Logger()
+
+	_, ok := vs.Annotations[VSAnnotationRateLimitPolicy]
+	if !ok {
+		rlpNamespacedName := client.ObjectKeyFromObject(rlp)
+		vs.Annotations[VSAnnotationRateLimitPolicy] = rlpNamespacedName.String()
+		if err := r.Client().Update(ctx, vs); err != nil {
```
eguzki, line 144: There needs to be a good reason to update a resource that the controller is not owner of. 

What is the purpose of adding that annotation?

controllers/apim/ratelimitpolicy_controller.go
```diff
@@ -85,15 +100,343 @@ func (r *RateLimitPolicyReconciler) Reconcile(eventCtx context.Context, req ctrl
...
+			logger.Error(err, "failed to add ratelimitpolicy annotation to virtualservice")
+			return err
+		}
+		logger.V(1).Info("successfully added ratelimitpolicy annotation to virtualservice")
+	}
+
+	// TODO(rahulanand16nov): store context of virtualservice in RLP's status block and manage envoy patches.
+	if err := r.reconcileRateLimitFilters(ctx, vs, rlp); err != nil {
```
eguzki, line 152: envoyfilters reconciliation in a method called "reconcileWithVirtualService` feels weird

controllers/apim/ratelimitpolicy_controller.go
```diff
@@ -85,15 +100,343 @@ func (r *RateLimitPolicyReconciler) Reconcile(eventCtx context.Context, req ctrl
...
+							StringValue: PatchedLimitadorClusterName,
+						},
+					},
+				},
+			},
+		},
+	}
+
+	factory := istioprovider.EnvoyFilterFactory{
+		ObjectName: gateway.Name + "-ratelimit-filters",
```
eguzki, line 335: I wonder if the envoyfilter resource name is linked to the rate limiting policy, more than  to the gateway

eguzki (changes requested):
Related to the ratelimitpolicy API, Currently it is defined with `routes.name`. Is this a better option that based on rules (like authorization policy)?

In this [gist](https://gist.github.com/eguzki/b9cabed957760fc03142dc9d85419ba5), there are rules like:
```yaml
spec:
  phases: ["postAuth"]
  rules:  # All conditions inside a single match block have AND semantics while match blocks have OR semantics
    - hosts:  # optional
      - toystore.example.com
      methods:  # optional
      - POST
      paths:  # optional
      - "/admin/addToy"
      notPaths: []  # optional
      notHosts: []  # optional
      notMethods: []  # optional
```

Obviously it is more complicated to "match" rules with the VS routes. However it is more aligned to reusable istio policies. what are your thoughts on this?

maleck13:
> Related to the ratelimitpolicy API, Currently it is defined with `routes.name`. Is this a better option that based on rules (like authorization policy)?
> 
> In this [gist](https://gist.github.com/eguzki/b9cabed957760fc03142dc9d85419ba5), there are rules like:
> 
> ```yaml
> spec:
>   phases: ["postAuth"]
>   rules:  # All conditions inside a single match block have AND semantics while match blocks have OR semantics
>     - hosts:  # optional
>       - toystore.example.com
>       methods:  # optional
>       - POST
>       paths:  # optional
>       - "/admin/addToy"
>       notPaths: []  # optional
>       notHosts: []  # optional
>       notMethods: []  # optional
> ```
> 
> Obviously it is more complicated to "match" rules with the VS routes. However it is more aligned to reusable istio policies. what are your thoughts on this?

Currently the routematch in the envoy filter object only supports ```name``` and ```action``` so I think for this to work with the API you are proposing we would have to go the route of WASM?

maleck13:
> Related to the ratelimitpolicy API, Currently it is defined with `routes.name`. Is this a better option that based on rules (like authorization policy)?
> 
> In this [gist](https://gist.github.com/eguzki/b9cabed957760fc03142dc9d85419ba5), there are rules like:
> 
> ```yaml
> spec:
>   phases: ["postAuth"]
>   rules:  # All conditions inside a single match block have AND semantics while match blocks have OR semantics
>     - hosts:  # optional
>       - toystore.example.com
>       methods:  # optional
>       - POST
>       paths:  # optional
>       - "/admin/addToy"
>       notPaths: []  # optional
>       notHosts: []  # optional
>       notMethods: []  # optional
> ```
> 
> Obviously it is more complicated to "match" rules with the VS routes. However it is more aligned to reusable istio policies. what are your thoughts on this?

Currently the routematch in the envoy filter object only supports name and action so I think for this to work with the API you are proposing we would have to go the route of WASM?

controllers/apim/ratelimitpolicy_controller.go
maleck13: perhaps we should create a TODO or issue to tack the need to change this?

controllers/apim/ratelimitpolicy_controller.go
```diff
@@ -85,15 +100,343 @@ func (r *RateLimitPolicyReconciler) Reconcile(eventCtx context.Context, req ctrl
...
+			},
+		},
+		Patch: &prePatch,
+	}
+
+	postAuthFilterPatch := preAuthFilterPatch
+	postAuthFilterPatch.Patch = postPatch
+
+	// Eventually, this should be dropped since it's a temp-fix for Kuadrant/limitador#53
+	clusterPatch := networkingv1alpha3.EnvoyFilter_EnvoyConfigObjectPatch{
```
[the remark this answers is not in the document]
rahulanand16nov: Not aware of that patch and difference doesn't matter that match except `clusterEnvoyPatch` adds a new cluster but the one in the PR changes the actual name to `rate_limit_cluster`. Will reuse `clusterEnvoyPatch`.

controllers/apim/ratelimitpolicy_controller.go
```diff
@@ -85,15 +100,343 @@ func (r *RateLimitPolicyReconciler) Reconcile(eventCtx context.Context, req ctrl
...
+			}
+
+			vHostName := host + ":80" // Istio names virtual host in this format: host + port
+			routePatch := routeRateLimitsPatch(vHostName, route.Name, route.Actions, route.Stage)
+
+			if err := controllerutil.SetOwnerReference(rlp, routePatch, r.Client().Scheme()); err != nil {
+				logger.Error(err, "failed to add owner ref to route-level ratelimits patch")
+				return err
+			}
+			if err := r.Client().Create(ctx, routePatch); err != nil {
```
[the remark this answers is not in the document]
rahulanand16nov: same [reply](https://github.com/Kuadrant/kuadrant-controller/pull/76/files#r789825831)

controllers/apim/ratelimitpolicy_controller.go
maleck13: Yes not sure this is needed now that we have changed the spec to have a targetRef. Currently the rate limit policy looks in the same namespace. So we should be able to put a watch on VirtualServices and HTTPRoutes and when one changes in a way we need to do something about, look up any RLP in same namespace and identify any that target the changed resource.

controllers/apim/ratelimitpolicy_controller.go
rahulanand16nov: For some reason, deserializing actions into that unstructured patch was working properly but I agree this needs a better way. Will create an issue when the PR is merged.

controllers/apim/ratelimitpolicy_controller.go
rahulanand16nov: It provided a mechanism to map to a particular RLP when VirtualService changes and is being watched. Considering what Craig said, this can be removed and I'll update the PR.

controllers/apim/ratelimitpolicy_controller.go
rahulanand16nov: The idea is we have two networking sources, namely, VirtualService and HTTPRoute. So, when VS changes, use `reconcileWithVirtualService` and so forth for HTTPRoute. Reconciling of envoyfilters is required since they rely on gateway names from networking sources.

maleck13:
tried this out locally. Added new rate limit rule for the get descriptor all worked as expected. I think when we demo this it would be good to show 

1) A global rate limit for all end points
2) a non authenticated rate limit for an endpoint
3) an authenticated rate limit for the admin endpoint

WDYT?

controllers/apim/ratelimitpolicy_controller.go
rahulanand16nov: ratelimit filters patch is required once per gateway so if it's per RLP then we'll have redundant patches to the same gateway if multiple RLPs target same gateway.

controllers/apim/ratelimitpolicy_controller.go
rahulanand16nov: Yes, it feels a bit weird to me as well. I'll change this as well :thinking:

rahulanand16nov:
I think, we can move the route patches into a single EnvoyFilter resource and hence management of them. Will try that as well :)

controllers/apim/ratelimitpolicy_controller.go
maleck13: Do we want to remove this as part of this PR?

maleck13:
@eguzki are you happy with the PR in its current state ?
@rahulanand16nov I think we should capture any outstanding items in Jira that we don't intend to fix here so that we do not lose track of them. This could be a simple ```Harden RateLimitPolicy reconciler ``` jira with a set of bullet points need to complete the work

rahulanand16nov:
> @eguzki are you happy with the PR in its current state ? @rahulanand16nov I think we should capture any outstanding items in Jira that we don't intend to fix here so that we do not lose track of them. This could be a simple `Harden RateLimitPolicy reconciler ` jira with a set of bullet points need to complete the work

OOO for today, I have some changes planned that I'll push tomorrow and add rest to a JIRA. Let's wait till tomorrow and we can merge this PR and move focus to auth.

rahulanand16nov:
I'll merge this for now and move the remaining details into the existing spike JIRA. Moving the focus to auth.

controllers/apim/ratelimitpolicy_controller.go
rahulanand16nov: Removing this and watching VS/HR is a relatively big change. I'll add them in the next PR.

PR #77, opened 2022-01-20
Fix authorino reference

eguzki:
Upgrading to authorino v0.7.0 (deployed by operator v0.1.0) raised some issues:

* group version changed from `authorino.3scale.net` to `authorino.kuadrant.io`
* Authorino operator deployed service name changed, breaking kuadrant deployment (`uthorino-authorization.kuadrant-system.svc.cluster.local` -> `authorino-authorino-authorization.kuadrant-system.svc.cluster.local`

Additionally, some changes were included:
* Istio upgrade to 1.12.1
* Testing tools using k8s 1.23.1

This PR is meant to have a (last) working version (v0.2.1) of the kuadrant controller with the deprecated API based on APIProduct / API


utils/local-deployment/patches/istio-externalProvider.yaml
```diff
@@ -5,7 +5,7 @@ spec:
     extensionProviders:
       - name: "kuadrant-authorization"
         envoyExtAuthzGrpc:
-          service: "authorino-authorization.kuadrant-system.svc.cluster.local"
+          service: "authorino-authorino-authorization.kuadrant-system.svc.cluster.local"
```
guicassolato, line 8: Notice that the `authorino-` prefix in the name of service is determined by the name of the `Authorino` CR – i.e. the name of the service is `{cr.name}-authorino-authorization`.

I understand it might be hard to couple what goes in this line to what's in `utils/local-deployment/authorino.yaml` and avoid it to break in case one changes either one or the other independently (at least while not using something like kustomize variables). Therefore, I don't expect you to change anything.

Just raising a flag that this is somehow fragile.

guicassolato (approved):

utils/local-deployment/patches/istio-externalProvider.yaml
eguzki: Indeed. 

The content in that patch ends up in the istio manifests in `utils/local-deployment/istio-manifests/Base/Pilot/Pilot.yaml`

Actually in this repo, this is all about dev/testing environment, so not a big deal. But this issue raises a flag for the kuadrant operator. The kuadrant operator will not deploy istio, but needs to add some configuration to it in order to register authorino as external authz provider. cc @mikenairn 

For this repo, I can add a CI test to make sure authorino service name is consistent in both places.

mikenairn: @eguzki Been thinking about this, do you think this would even be possible from another operator? 

If I'm understanding this correctly, we need to add the configuration for kuadrant defined in that patch file to the istio config. However, this is an admin task for the istio installation itself, and we(kuadrant-operator) won't necessarily know anything about that deployment on the cluster, or have access to it.

We can make the kuadrant operator work by installing istio with the extension provider config already added, like you are doing here, before installing the operator, but that is obviously not a solution since we want it to work with an already installed istio. Does istio provide any other way to customise istio config? CR?

eguzki: Indeed, this permission issue needs to be explored. I commented [here](https://issues.redhat.com/browse/KUADRANT-13?focusedCommentId=19632696&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-19632696)

Istio Operator CR can do the work. However, a not istio admin can do this?? to be explored:

```yaml
apiVersion: install.istio.io/v1alpha1
kind: IstioOperator
spec:
  meshConfig:
    extensionProviders:
      - name: "kuadrant-authorization"
        envoyExtAuthzGrpc:
          service: "authorino-authorino-authorization.kuadrant-system.svc.cluster.local"
          port: 50051
  components:
    ingressGateways:
      - name: kuadrant-gateway
        enabled: true
        namespace: kuadrant-system
        label:
          istio: kuadrant-system
```

PR #78, opened 2022-01-27
Update getting started doc

mikenairn:
Update the expected pods/deployments after kuadrant installation.

### DIFF
--- a/doc/getting-started.md
+++ b/doc/getting-started.md
@@ -52,7 +52,7 @@ On successful command return, you should see the following deployments and pods 
 ```bash
 ❯ kubectl get pods -n kuadrant-system
 NAME                                                     READY   STATUS    RESTARTS   AGE
-authorino-controller-manager-XXXXXXXXXXX-XXXX            2/2     Running   0          3m6s
+authorino-XXXXXXXXXXX-XXXX                               1/1     Running   0          2m33s
 istiod-XXXXXXXXXX-XXXXX                                  1/1     Running   0          3m11s
 kuadrant-controller-manager-XXXXXXXXXX-XXXX              2/2     Running   0          3m5s
 kuadrant-gateway-XXXXXXXXXX-XXXX                         1/1     Running   0          3m5s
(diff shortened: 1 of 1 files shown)


doc/getting-started.md
```diff
@@ -62,12 +62,23 @@ limitador-operator-controller-manager-XXXXXXXXXX-XXXXX   2/2     Running   0
...
+
+
+❯ kubectl get pods -n authorino-operator
+NAME                                                     READY   STATUS    RESTARTS   AGE
+authorino-operator-controller-manager-XXXXXXXXXXX-XXXX   2/2     Running   0          2m36s
+
+
+❯ kubectl get deployments -n authorino-operator
+NAME                                    READY   UP-TO-DATE   AVAILABLE   AGE
+authorino-operator-controller-manager   1/1     1            1           2m36s
```
mikenairn, line 80: Not sure if this is intentional, but the authorino operator is deployed in it's own namespace and not `kuadrant-system`.

guicassolato: It's expected when deployed using https://github.com/Kuadrant/authorino-operator/blob/bb5ad6be30ec220cd9e99875cacdb9b767317d41/config/deploy/manifests.yaml#L2038

mikenairn: It looks like kuadrantctl was changed to just pull those generated manifests, so i guess that's what will happen now, but is that what we want here? Should the kuadrant installation not put all the kuadrant pieces in the one ns?

eguzki: I asked about easy way to kustomize manifests to deploy authorino operator in `kuadrant-system`. But it was not easy (or straightforward), so the kuadrantctl tool installs it in default namespace, i.e. `authorino-operator` ns. 

About the appropriate place to be deployed... I also think it should be `kuadrant-system`.. but it is not a big deal also.

mikenairn: You could probably setup something similar to what i did here in the [kuadrant operator](https://github.com/Kuadrant/kuadrant-operator/blob/main/config/dependencies/kustomization.yaml). Adds a kustomizaion for each component that just points to the components own [kustomization directory on github](https://github.com/Kuadrant/kuadrant-operator/blob/main/config/dependencies/authorino/kustomization.yaml#L2), you can then change the namespace of all the resources in the usual way like [here](https://github.com/Kuadrant/kuadrant-operator/blob/main/config/dependencies/kustomization.yaml#L1)

You may or may not need to delete namespace resources if there is more than one being added along the way, but you can do that like [this](https://github.com/Kuadrant/kuadrant-operator/blob/main/config/dependencies/kustomization.yaml#L10) if needs be. 

Your manifest generation make targets would end up like:

```
$(KUSTOMIZE) build config/component/authorino > /authorinomanifests/autogenerated/authorino-operator.yaml
```

If your referencing the kustomizations from a tagged release there should be no need to manipulate the manifest any more than changing the namespace i would have thought.

Regardless, do we want to merge this PR to update the docs since they are currently wrong for the current kuadrantctl release.

eguzki: Actually, most of the doc was based on the APIProduct/API CRD model. So when we remove it, doc needs to be changed entirely

PR #79, opened 2022-01-28
create AuthorizationPolicy using VirtualService

rahulanand16nov:
This PR:
- Creates AuthorizationPolicy using VirtualService and an annotation stating the CUSTOM extension provider.
- Deletes the policy if VirtualService is deleted or Annotation is removed.

Notes:
- Regex ([StringMatch](https://istio.io/latest/docs/reference/config/networking/virtual-service/#StringMatch)) is not supportable as AuthPolicy only supports: exact, prefix, and suffix. So, right now, only taking care of exact and prefix StringMatch types.
- Annotation changed from `kuadrant.io/auth-enabled: true` to `kuadrant.io/auth-provider: "sample-extz-auth-provider"`

## Steps to Verify
1. Add extension provider into the mesh configuration. https://istio.io/latest/docs/tasks/security/authorization/authz-custom/
3. `make install && make run`
4. Add `auth-provider` annotation to the virtualservice (`examples/toystore/virtualservice.yaml`)
5. Verify generated authorizationPolicy

### DIFF
--- a/controllers/apim/virtualservice_controller.go
+++ b/controllers/apim/virtualservice_controller.go
@@ -1,0 +1,216 @@
+package apim
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	networkingv1alpha3 "istio.io/api/networking/v1alpha3"
+	securityv1beta1 "istio.io/api/security/v1beta1"
+	istionetworkingv1alpha3 "istio.io/client-go/pkg/apis/networking/v1alpha3"
+	istiosecurityv1beta1 "istio.io/client-go/pkg/apis/security/v1beta1"
+
+	"github.com/go-logr/logr"
+	"github.com/kuadrant/kuadrant-controller/pkg/log"
+	"github.com/kuadrant/kuadrant-controller/pkg/reconcilers"
+)
+
+const (
+	KuadrantAuthProviderAnnotation = "kuadrant.io/auth-provider"
+)
+
+// VirtualServiceReconciler reconciles Istio's AuthorizationPolicy object
+type VirtualServiceReconciler struct {
+	*reconcilers.BaseReconciler
+	Scheme *runtime.Scheme
+}
+
+func (r *VirtualServiceReconciler) Reconcile(eventCtx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := r.Logger().WithValues("VirtualService", req.NamespacedName)
+	ctx := logr.NewContext(eventCtx, logger)
+
+	virtualService := istionetworkingv1alpha3.VirtualService{}
+	if err := r.Client().Get(ctx, req.NamespacedName, &virtualService); err != nil {
+		if apierrors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		logger.Error(err, "failed to get VirtualService")
+		return ctrl.Result{}, err
+	}
+
+	if err := r.reconcileAuthPolicy(ctx, &virtualService); err != nil {
+		logger.Error(err, "failed to reconcile AuthorizationPolicy")
+		return ctrl.Result{}, err
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func getAuthPolicyName(vsName string) string {
+	return "source-virtualservice-" + vsName
+}
+
+func (r *VirtualServiceReconciler) reconcileAuthPolicy(ctx context.Context, vs *istionetworkingv1alpha3.VirtualService) error {
+	logger := r.Logger()
+	logger.Info("Reconciling AuthorizationPolicy")
+
+	authPolicy := istiosecurityv1beta1.AuthorizationPolicy{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      getAuthPolicyName(vs.Name),
+			Namespace: vs.Namespace,
+		},
+	}
+
+	providerName, present := vs.GetAnnotations()[KuadrantAuthProviderAnnotation]
+	if !present {
+		return fmt.Errorf("Kuadrant auth-provider annotation not found")
+	}
+
+	// fill out the rules
+	authToRules := []*securityv1beta1.Rule_To{}
+	for _, httpRoute := range vs.Spec.Http {
+		for idx, matchRequest := range httpRoute.Match {
+			toRule := &securityv1beta1.Rule_To{
+				Operation: &securityv1beta1.Operation{},
+			}
+
+			toRule.Operation.Hosts = vs.Spec.Hosts
+			if normalizedURI := normalizeStringMatch(matchRequest.Uri); normalizedURI != "" {
+				toRule.Operation.Paths = append(toRule.Operation.Paths, normalizedURI)
+			}
+
+			if normalizedMethod := normalizeStringMatch(matchRequest.Method); normalizedMethod != "" {
+				// Looks like it's case-sensitive:
+				// https://istio.io/latest/docs/reference/config/security/normalization/#1-method-not-in-upper-case
+				method := strings.ToUpper(normalizedMethod)
+				toRule.Operation.Methods = append(toRule.Operation.Methods, method)
+			}
+
+			// If there is only regex stringmatches then we'll have bunch of repeated To rules with
+			// only same host filled into each. Following make sure only one field like that is present.
+			operation := toRule.Operation
+			if len(operation.Paths) == 0 && len(operation.Methods) == 0 && idx > 0 {
+				continue
+			}
+			authToRules = append(authToRules, toRule)
+		}
+	}
+
+	authPolicy.Spec = securityv1beta1.AuthorizationPolicy{
+		Rules: []*securityv1beta1.Rule{{
+			To: authToRules,
+		}},
+		Action: securityv1beta1.AuthorizationPolicy_CUSTOM,
+		ActionDetail: &securityv1beta1.AuthorizationPolicy_Provider{
+			Provider: &securityv1beta1.AuthorizationPolicy_ExtensionProvider{
+				Name: providerName,
+			},
+		},
+	}
+
+	if err := controllerutil.SetOwnerReference(vs, &authPolicy, r.Client().Scheme()); err != nil {
+		logger.Error(err, "failed to add owner ref to AuthorizationPolicy resource")
+		return err
+	}
+	err := r.ReconcileResource(ctx, &istiosecurityv1beta1.AuthorizationPolicy{}, &authPolicy, alwaysUpdateAuthPolicy)
+	if err != nil && !apierrors.IsAlreadyExists(err) {
+		logger.Error(err, "ReconcileResource failed to create/update AuthorizationPolicy resource")
+		return err
+	}
+
+	logger.Info("successfully created/updated AuthorizationPolicy resources")
+	return nil
+}
+
+func alwaysUpdateAuthPolicy(existingObj, desiredObj client.Object) (bool, error) {
+	existing, ok := existingObj.(*istiosecurityv1beta1.AuthorizationPolicy)
+	if !ok {
+		return false, fmt.Errorf("%T is not a *istiosecurityv1beta1.AuthorizationPolicy", existingObj)
+	}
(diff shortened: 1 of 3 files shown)


controllers/apim/virtualservice_controller.go
```diff
@@ -0,0 +1,216 @@
...
+	authPolicy := istiosecurityv1beta1.AuthorizationPolicy{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      getAuthPolicyName(vs.Name),
+			Namespace: vs.Namespace,
+		},
+	}
+
+	providerName, present := vs.GetAnnotations()[KuadrantAuthProviderAnnotation]
+	if !present {
+		return fmt.Errorf("Kuadrant auth-provider annotation not found")
```
maleck13, line 78: I don't think this needs to be an error? Perhaps we just log and return?

rahulanand16nov: Yea, we can just log it. updated.

controllers/apim/virtualservice_controller.go
```diff
@@ -0,0 +1,216 @@
...
+				toRule.Operation.Methods = append(toRule.Operation.Methods, method)
+			}
+
+			// If there is only regex stringmatches then we'll have bunch of repeated To rules with
+			// only same host filled into each. Following make sure only one field like that is present.
+			operation := toRule.Operation
+			if len(operation.Paths) == 0 && len(operation.Methods) == 0 && idx > 0 {
+				continue
+			}
+			authToRules = append(authToRules, toRule)
```
maleck13, line 109: this is controlling which paths and methods go to Authorino? Do we need that level or for now should we just send any request to the host?

rahulanand16nov: yes, specific paths. Better to specific I guess?

controllers/apim/virtualservice_controller.go
```diff
@@ -0,0 +1,216 @@
...
+		return false, fmt.Errorf("%T is not a *istiosecurityv1beta1.AuthorizationPolicy", desiredObj)
+	}
+
+	existing.Spec = desired.Spec
+	return true, nil
+}
+
+func normalizeStringMatch(sm *networkingv1alpha3.StringMatch) string {
+	if prefix := sm.GetPrefix(); prefix != "" {
+		return prefix + "*"
```
maleck13, line 155: why do we need to add this? Is it due to a difference between VS and AuthorizationPolicy? Does prefix always end with ```/``` or is it valid to have ```/cars*```?

rahulanand16nov: Yes, due to differences between them. From the docs `Prefix match: “abc*” will match on value “abc” and “abcd”.`, it looks like it will be valid.

controllers/apim/virtualservice_controller.go
```diff
@@ -0,0 +1,216 @@
...
+		if err := m.K8sClient.Get(context.Background(), authObjKey, authPolicy); err != nil {
+			// no annotation but authpolicy exist means annotation was removed
+			if !apierrors.IsNotFound(err) {
+				m.Logger.Error(err, "failed to check AuthorizationPolicy existence")
+			}
+			return nil // this virtualservice is not protected
+		}
+
+		// AuthorizationPolicy exists
+		if err := m.K8sClient.Delete(context.Background(), authPolicy); err != nil {
```
maleck13, line 190: this feels weird in a map function should we not just return the reconcile request and handle this in the reconcile logic?

rahulanand16nov: Creating a request for it means creating for all the events. Shouldn't we only care about virtualservices with annotation in the reconcile logic?

rahulanand16nov: I can move it if you want?

eguzki (changes requested):

controllers/apim/virtualservice_controller.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *VirtualServiceReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	virtualServiceFilter := VirtualServiceFilter{
+		K8sClient: r.Client(),
+		Logger:    r.Logger().WithName("AuthPolicyHandler"),
+	}
+	return ctrl.NewControllerManagedBy(mgr).
+		Watches(
```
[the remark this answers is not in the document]
eguzki: Check the services controller of the current main branch. Use a label instead of annotation for builtin filtering out of events. https://github.com/Kuadrant/kuadrant-controller/blob/main/controllers/service_controller.go#L341

```go
return ctrl.NewControllerManagedBy(mgr).                                                                   
    For(&corev1.Service{}, builder.WithPredicates(kuadrantServicesPredicate())).                           
    Owns(&networkingv1beta1.API{}).                                                                        
    WithLogger(log.Log). // use base logger, the manager will add prefixes for watched sources             
    Complete(r) 
```

eguzki (changes requested):

controllers/apim/virtualservice_controller.go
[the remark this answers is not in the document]
rahulanand16nov: `For` and `Owns`, confused me a bit. But it makes sense since AuthorizationPolicy is being generated. But I am not sure we need that `WithPredicated` since filtering is the same for all the events.

controllers/apim/virtualservice_controller.go
rahulanand16nov: This behavior is similar to AuthPolicy generated using OAS.

controllers/apim/virtualservice_controller.go
guicassolato: I agree with @maleck13 . It looks like this is changing the state of the cluster while filtering the resources.

guicassolato:
About step number 1 to verify:

> Add extension provider into the mesh configuration. https://istio.io/latest/docs/tasks/security/authorization/authz-custom/

If you deploy with https://github.com/Kuadrant/kuadrantctl, an extension provider should be defined with name `kuadrant-authorization` already:

https://github.com/Kuadrant/kuadrantctl/blob/578161bea609a0ed053e2ce6f457911a257a10b6/istiomanifests/autogenerated/Base/Pilot/Pilot.yaml#L190

controllers/apim/virtualservice_controller.go
rahulanand16nov: Filter is now done in the reconcile logic along with the orphan check.

controllers/apim/virtualservice_controller.go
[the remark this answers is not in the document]
rahulanand16nov: Thanks, @eguzki! looks much more simpler now with `For` and `Owns` :)

rahulanand16nov:
@guicassolato so it's in fact automatically added, nice!

Sample of generated AuthorizationPolicy:
```yaml
apiVersion: security.istio.io/v1beta1
  kind: AuthorizationPolicy
  metadata:
    creationTimestamp: "2022-01-31T12:26:56Z"
    generation: 1
    name: source-virtualservice-toystore
    namespace: kuadrant-system
    ownerReferences:
    - apiVersion: networking.istio.io/v1alpha3
      kind: VirtualService
      name: toystore
      uid: 4ce46ede-9775-4a16-a422-e3297f9018db
    resourceVersion: "5523"
    uid: a2badcc7-b1f3-4298-b1ce-d6ec56b55805
  spec:
    action: CUSTOM
    provider:
      name: sample-extz-auth-provider
    rules:
    - to:
      - operation:
          hosts:
          - '*.toystore.com'
          methods:
          - GET
          paths:
          - /getToy
      - operation:
          hosts:
          - '*.toystore.com'
          methods:
          - POST
          paths:
          - /admin/addToy
```

controllers/apim/virtualservice_controller.go
rahulanand16nov: I remembered that I added the ownerRef to the AuthPolicy so, in theory, orphan authpolicy shouldn't exist at all. I'll just remove that check :thinking:

controllers/apim/virtualservice_controller.go
```diff
@@ -0,0 +1,218 @@
...
+
+	return ctrl.Result{}, nil
+}
+
+func getAuthPolicyName(vsName string) string {
+	return "source-virtualservice-" + vsName
+}
+
+func (r *VirtualServiceReconciler) reconcileAuthPolicy(ctx context.Context, vs *istionetworkingv1alpha3.VirtualService) error {
+	logger := r.Logger()
```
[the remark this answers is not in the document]
rahulanand16nov: Hmm, I see :thinking:.... Updated.

controllers/apim/virtualservice_controller.go
rahulanand16nov: Done.

rahulanand16nov: Sorry, my bad, I added orphan check because let's say the user removes the annotation, and there is no automatic removal of AuthPolicy. I re-added the check.

rahulanand16nov:
@maleck13 @eguzki @guicassolato Approvals before I merge?

maleck13 (commented):
This looks pretty good to me @eguzki if you are happy lets go ahead and merge

eguzki:
I do not want to block the PR any longer, so it gets my LGTM.

However, there is  one improvement  I wanted to highlight.

Filter events to reconcile only Virtualservice objects with kuadrant labels. Currently, any change on any virtualservice of the cluster will generate reconciliation loop and at least two GET operations against the k8s API server. Adding predicates (like the [service controller implemented here](https://github.com/Kuadrant/kuadrant-controller/blob/main/controllers/service_controller.go#L343))  to the watcher not only avoids unwanted reconciliation loops, also makes the code easier as you can assume the event is coming from a kuadrant enabled virtual service. And if it does not have the label, you know they just removed it and your duty is to delete the managed authorizationpolicy.

eguzki (approved):

rahulanand16nov:
> I do not want to block the PR any longer, so it gets my LGTM.
> 
> However, there is one improvement I wanted to highlight.
> 
> Filter events to reconcile only Virtualservice objects with kuadrant labels. Currently, any change on any virtualservice of the cluster will generate reconciliation loop and at least two GET operations against the k8s API server. Adding predicates (like the [service controller implemented here](https://github.com/Kuadrant/kuadrant-controller/blob/main/controllers/service_controller.go#L343)) to the watcher not only avoids unwanted reconciliation loops, also makes the code easier as you can assume the event is coming from a kuadrant enabled virtual service. And if it does not have the label, you know they just removed it and your duty is to delete the managed authorizationpolicy.

@eguzki I wanted to implement that but was not sure how to delete the AuthorizationPolicy. When users delete the label, how can the controller know which AuthPolicy to delete if it doesn't reach the Reconcile logic (filtered before enqueuing)?

eguzki:
Deletion should not be done in the mapper. I agree. But, if you look at the predicate implementation linked previously, the Update function checks for the label on the `old` object. Even if the label was deleted, the predicate tells the manager to get the event and run reconciliation event. In the reconciler,  check for the existence of the label. If it is not there, you know it used to be there but not anymore, hence proceed with the deletion of the owned object. If you happen to know de name of the authorizationpolicy, even easier, just delete the object.

guicassolato (commented):
Tested locally, with the kuadrant-controller image built and pushed to the cluster (instead of `make run`), rolled out on top of the default stable image deployed by `kuadrantctl install`.

I didn't have to define the extension provider into the istio configmap  as it's set by default by kuadrantctl. All I had to do was:
1. apply `examples/toystore/toystore.yaml`
2. apply `examples/toystore/virtualservice.yaml` with added annotation `kuadrant.io/auth-provider: kuadrant-authorization`
3. create an `AuthConfig` + API key secret for host `api.toystore.com`
4. `kubectl -n kuadrant-system port-forward services/kuadrant-gateway 8000:80 &`
5. `curl -H 'Host: api.toystore.com' -H 'Authorization: APIKEY ...' http://localhost:8000/getToy # 200`

Then I tried:
- `kubectl -n kuadrant-system delete virtualservices/toystore` → deletes the `AuthorizationPolicy` and requests to the API start returning 404
- Removing the annotation from the virtual service → deletes the `AuthorizationPolicy` but requests to the API continue to be served – i.e. the kuadrant-gateway continues to check with Authorino, which still knows the host given the `AuthConfig` is still there.

rahulanand16nov:
@eguzki Added predicates.

@guicassolato Looks expected behavior except the last `the kuadrant-gateway continues to check with Authorino, which still knows the host given the AuthConfig is still there.`, right? But I don't feel it's a problem with this PR?

rahulanand16nov:
Thanks guys for the amazing feedback! Looks good enough to merge for now.

rahulanand16nov:
Reference AuthConfig and secret for easy testing for the future:
```
kubectl -n kuadrant-system apply -f -<<EOF
apiVersion: authorino.kuadrant.io/v1beta1
kind: AuthConfig
metadata:
  name: simple-api-protection
spec:
  hosts:
  - api.toystore.io
  identity:
  - name: friends
    apiKey:
      labelSelectors:
        group: friends
    credentials:
      in: authorization_header
      keySelector: APIKEY
EOF
```
```
kubectl -n kuadrant-system apply -f -<<EOF
apiVersion: v1
kind: Secret
metadata:
  name: api-key
  labels:
    authorino.kuadrant.io/managed-by: authorino
    group: friends
stringData:
  api_key: SUPER_SECRET_STUFF
type: Opaque
EOF
```

PR #80, opened 2022-02-01
Fixate preauth as the first and postauth as the last ratelimit filter

rahulanand16nov:
To make unauthenticated rate-limit and authenticated rate-limit work, we have to make sure the relative orders of four filters is as follows:
```
                              ┌─────PreAuth-RateLimit-Filter
                              │
                              │     RBAC-Filter  ─────────────────────┐
Added by RateLimitPolicy ─────┤                                       ├───── Added by AuthorizationPolicy
                              │     Ext-Authorization-Filter ─────────┘
                              │
                              └─────PostAuth-RateLimit-Filter
```

The best way to make sure the order stays like above is to insert preauth with `INSERT_FIRST` (just under HTTP.connection.manager) and postauth with `INSERT_BEFORE` (router) operations in EnvoyFIlter. This makes sure that in whichever order you apply AuthPolicy and RLPolicy, you'll get the same result.

## Verification Steps
1. `kuadrantctl install`
2. `make install && make run`
3. `kubectl apply -f examples/toystore/toystore.yaml`
4. `kubectl apply -f examples/toystore/virtualService.yaml` + add annotation (`kuadrant.io/auth-provider: kuadrant-authorization`) // This will generate AuthorizationPolicy
5. `kubectl apply -f config/samples/apim_v1alpha1_ratelimitpolicy.yaml`
6. `istioctl dashboard envoy <kuadrant-gateway-pod>.kuadrant-system` //`config_dump` and search `preauth`

You can swap 4 and 5 to get the same result in step 6.

### DIFF
--- a/config/samples/apim_v1alpha1_ratelimitpolicy.yaml
+++ b/config/samples/apim_v1alpha1_ratelimitpolicy.yaml
@@ -8,7 +8,7 @@ metadata:
 spec:
   networkingRef:
     - type: VirtualService
-      name: toystore-virtual-service
+      name: toystore
   routes:
     - name: get-toy
       stage: BOTH
--- a/controllers/apim/ratelimitpolicy_controller.go
+++ b/controllers/apim/ratelimitpolicy_controller.go
@@ -45,6 +45,9 @@ const (
 	preAuthRLStage  = 0
 	postAuthRLStage = 1
 
+	EnvoysHTTPPortNumber            = 8080
+	EnvoysHTTPConnectionManagerName = "envoy.filters.network.http_connection_manager"
+
 	VSAnnotationRateLimitPolicy = "kuadrant.io/ratelimitpolicy"
 )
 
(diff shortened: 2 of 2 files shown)


eguzki:
question, would this filter order work?
```
   - RBAC-Filter
   - PreAuth-RateLimit-Filter
   - Ext-Authorization-Filter
   - PostAuth-RateLimit-Filter
```

rahulanand16nov:
> question, would this filter order work?
> 
> ```
>    - RBAC-Filter
>    - PreAuth-RateLimit-Filter
>    - Ext-Authorization-Filter
>    - PostAuth-RateLimit-Filter
> ```

Yup, it would. Thing is Istio adds RBAC and ext-authz at the same time.

guicassolato:
Why does the RBAC filter have to be added? Can we avoid that since RBAC checks are supposed to be modelled inside the `AuthConfig`?

rahulanand16nov:
> Why does the RBAC filter have to be added? Can we avoid that since RBAC checks are supposed to be modelled inside the `AuthConfig`?

It's added by Istio along with the ext-auth filter so no control over avoiding its addition I guess. RBAC filter's policy is generated from the information of AuthorizationPolicy's rules. It helps the envoy decide for which requests should it reach out to an external authorizer (Authorino).

eguzki (approved):

rahulanand16nov:
@maleck13 found a bug. For some reason, if you try the same steps but with `make local-setup` preauth filter is not added to the filter chain.

maleck13:
ok based on that I am happy to merge at this point

rahulanand16nov:
Before the merge, I had rebased `new-api-impl` and `fix-filters-order` on top of master. It's working fine as expected. Cheers :)

PR #82, opened 2022-02-07
Harden RateLimitPolicy implementation

rahulanand16nov:
This PR adds the following changes:
1. Adds uniqueness requirement on networking refs. This means you cannot add duplicate networking refs of the same type but can with different types (VirtualService and HTTPRoute).
2. Registers Istio types because they were removed in the last PR and caused the controller to stop working.
3. Moves away from using `ownerRefs` on EnvoyFilter resources to using annotation to keep track of dependent RateLimitPolcies. This change was required because it was constraining us to only create resources in the gateway's namespace, which we don't want. This allows us to have VirtualService/RateLimitPolicy in any namespace, but EnvoyFilter patches will be created only in the relevant gateway's namespace.
4. Combines all the ratelimits patches into just one resource for easy management and simpler reconcile logic.
5. Passes labels of the gateways to the `EnvoyFilter` resources which allows flexibility to the constraints mentioned in the change `4`.
6. Uses finalizers to account for the `3` change. This means removing `ownerRLP` entries and if none are present, delete the patch itself.
7. Refactors for simplicity and code reuse.
8. Makes AuthPolicy per-gateway into the gateway's namespace.

### Initial steps
The following steps are common for all the verifying subsections.
1. `make local-setup`
3. `k apply -f examples/toystore/virtualService.yaml`

#### Verify change `1`
Following command should fail with the error: `The RateLimitPolicy "ratelimitpolicy" is invalid: spec.networkingRef[1]: Duplicate value: map[string]interface {}{"name":"toystore", "type":"VirtualService"}
`
```yaml
k apply -f -<<EOF
kind: RateLimitPolicy
metadata:
  name: ratelimitpolicy
spec:
  networkingRef:
    - type: VirtualService
      name: toystore
    - type: VirtualService
      name: toystore
EOF
```

#### Verify change `3`, `4` and `5`
```yaml
k apply -f -<<EOF
kind: RateLimitPolicy
metadata:
  name: ratelimitpolicy-1
spec:
  networkingRef:
    - type: VirtualService
      name: toystore
  routes:
    - name: add-toy
      stage: POSTAUTH
      actions:
        - generic_key:
            descriptor_key: admin
            descriptor_value: "yes"
  limits:
    - conditions: ["admin == yes"]
      max_value: 2
      namespace: postauth
      seconds: 15
      variables: []
EOF
```
```yaml
k apply -f -<<EOF
kind: RateLimitPolicy
metadata:
  name: ratelimitpolicy-2
spec:
  networkingRef:
    - type: VirtualService
      name: toystore
  routes:
    - name: get-toy
      stage: BOTH
      actions:
        - generic_key:
            descriptor_key: get
            descriptor_value: "yes"
  limits:
    - conditions: ["get == yes"]
      max_value: 2
      namespace: preauth
      seconds: 15
      variables: []
EOF
```
These commands create two `RateLimit` resources and three `EnvoyFilter` (one for filters and two for ratelimits) resources with `kuadrant.io/ownerRateLimitPolicies: default/ratelimitpolicy-1,default/ratelimitpolicy-2` annotation. Notice that those two values refer to the RLPs you just created. Plus, now there is only one ratelimit patch per RLP and they can be created even if RLP is in a different namespace to that of gateways.

#### Verify change `6`
This piggybacks onto the last verification section.
```
k delete ratelimitpolicy ratelimitpolicy-1
```
Check patches to see that the `ownerRLP` entry and one of ratelimits patch is removed.
```
k delete ratelimitpolicy ratelimitpolicy-2
```
This should remove all the `EnvoyFilter` resources that were created in the last section.

#### Verify change `8`
Add the `kuadrant.io/auth-provider: kuadrant-authorization` annotation to the VirtualService added in the initial step.
```
k get authorizationpolicy -A
```
This command will show the authpolicy in the gateway's namespace even though VirtualService is in `default` namespace.

### DIFF
--- a/apis/apim/v1alpha1/ratelimitpolicy_types.go
+++ b/apis/apim/v1alpha1/ratelimitpolicy_types.go
@@ -77,6 +77,9 @@ type NetworkingRef struct {
 
 // RateLimitPolicySpec defines the desired state of RateLimitPolicy
 type RateLimitPolicySpec struct {
+	//+listType=map
+	//+listMapKey=type
+	//+listMapKey=name
 	NetworkingRef []NetworkingRef `json:"networkingRef,omitempty"`
 	// route specific staging and actions
 	//+listType=map
--- a/controllers/apim/ratelimitpolicy_controller.go
+++ b/controllers/apim/ratelimitpolicy_controller.go
@@ -28,6 +28,7 @@ import (
 	networkingv1alpha3 "istio.io/api/networking/v1alpha3"
 	istio "istio.io/client-go/pkg/apis/networking/v1alpha3"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -39,12 +40,12 @@ import (
 )
 
 const (
-	finalizerName = "kuadrant.io/ratelimitpolicy"
-
 	EnvoysHTTPPortNumber            = 8080
 	EnvoysHTTPConnectionManagerName = "envoy.filters.network.http_connection_manager"
 
 	VSAnnotationRateLimitPolicy = "kuadrant.io/ratelimitpolicy"
+
+	InvalidNetworkingRefTypeErr = "unknown networking reference type"
 )
 
 // RateLimitPolicyReconciler reconciles a RateLimitPolicy object
@@ -56,6 +57,9 @@ type RateLimitPolicyReconciler struct {
 //+kubebuilder:rbac:groups=apim.kuadrant.io,resources=ratelimitpolicies,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=apim.kuadrant.io,resources=ratelimitpolicies/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=apim.kuadrant.io,resources=ratelimitpolicies/finalizers,verbs=update
+//+kubebuilder:rbac:groups=networking.istio.io,resources=gateways,verbs=get;list;watch
+//+kubebuilder:rbac:groups=networking.istio.io,resources=envoyfilters,verbs=get;list;watch;create;delete;update;patch
+//+kubebuilder:rbac:groups=limitador.kuadrant.io,resources=ratelimits,verbs=get;list;watch;create;update;delete;patch
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
@@ -74,23 +78,28 @@ func (r *RateLimitPolicyReconciler) Reconcile(eventCtx context.Context, req ctrl
 	var rlp apimv1alpha1.RateLimitPolicy
 	if err := r.Client().Get(ctx, req.NamespacedName, &rlp); err != nil {
 		if apierrors.IsNotFound(err) {
-			logger.Info("no rate limit policy found.")
+			logger.Info("no RateLimitPolicy found")
 			return ctrl.Result{}, nil
 		}
 		logger.Error(err, "failed to get RateLimitPolicy")
 		return ctrl.Result{}, err
 	}
 
-	if rlp.GetDeletionTimestamp() != nil && controllerutil.ContainsFinalizer(&rlp, finalizerName) {
-		controllerutil.RemoveFinalizer(&rlp, finalizerName)
+	if rlp.GetDeletionTimestamp() != nil && controllerutil.ContainsFinalizer(&rlp, filtersFinalizer) {
+		logger.Info("Removing finalizers")
+		if err := r.finalizeEnvoyFilters(ctx, &rlp); err != nil {
+			logger.Error(err, "failed to remove ownerRlp entry from filters patch")
+			return ctrl.Result{}, err
+		}
+		controllerutil.RemoveFinalizer(&rlp, filtersFinalizer)
 		if err := r.BaseReconciler.UpdateResource(ctx, &rlp); client.IgnoreNotFound(err) != nil {
-			return ctrl.Result{Requeue: true}, err
+			return ctrl.Result{}, err
 		}
 		return ctrl.Result{}, nil
 	}
 
-	if !controllerutil.ContainsFinalizer(&rlp, finalizerName) {
-		controllerutil.AddFinalizer(&rlp, finalizerName)
+	if !controllerutil.ContainsFinalizer(&rlp, filtersFinalizer) {
+		controllerutil.AddFinalizer(&rlp, filtersFinalizer)
 		if err := r.BaseReconciler.UpdateResource(ctx, &rlp); client.IgnoreNotFound(err) != nil {
 			return ctrl.Result{Requeue: true}, err
 		}
@@ -107,14 +116,14 @@ func (r *RateLimitPolicyReconciler) Reconcile(eventCtx context.Context, req ctrl
 			return ctrl.Result{}, nil
 		case apimv1alpha1.NetworkingRefType_VS:
 			vsNamespacedName := client.ObjectKey{
-				Namespace: rlp.Namespace,
+				Namespace: rlp.Namespace, // VirtualService lookup is limited to RLP's namespace
 				Name:      networkingRef.Name,
 			}
 
 			var vs istio.VirtualService
 			if err := r.Client().Get(ctx, vsNamespacedName, &vs); err != nil {
 				if apierrors.IsNotFound(err) {
-					logger.Info("no virtualservice found", "lookup name", vsNamespacedName.String())
+					logger.Info("no VirtualService found", "lookup name", vsNamespacedName.String())
 					return ctrl.Result{}, nil
 				}
 				logger.Error(err, "failed to get VirutalService")
@@ -122,11 +131,11 @@ func (r *RateLimitPolicyReconciler) Reconcile(eventCtx context.Context, req ctrl
 			}
 
 			if err := r.reconcileWithVirtualService(ctx, &vs, &rlp); err != nil {
-				logger.Error(err, "failed to reconcile with virtualservice")
+				logger.Error(err, "failed to reconcile with VirtualService")
 				return ctrl.Result{}, err
 			}
 		default:
-			err := fmt.Errorf("unknown networking reference type")
+			err := fmt.Errorf(InvalidNetworkingRefTypeErr)
 			logger.Error(err, "networking reconciliation failed")
 			return ctrl.Result{}, err
 		}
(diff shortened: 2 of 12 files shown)


controllers/apim/ratelimitpolicy_controller.go
```diff
@@ -136,100 +145,113 @@ func (r *RateLimitPolicyReconciler) Reconcile(eventCtx context.Context, req ctrl
...
 
-	if err := r.reconcileRateLimitActions(ctx, vs, rlp); err != nil {
-		return err
-	}
-	logger.Info("successfully reconciled ratelimitpolicy using attached virtualservice")
-	return nil
-}
+	// create/update EnvoyFilter patches for each gateway
+	for _, gw := range vs.Spec.Gateways {
+		gwKey := common.NamespacedNameToObjectKey(gw, vs.Namespace) // Istio defaults to VirtualService's namespace
```
maleck13, line 164: not sure I understand this. What is meant by Istio defaults to VirtualService's namespace?

controllers/apim/ratelimitpolicy_controller.go
```diff
@@ -136,100 +145,113 @@ func (r *RateLimitPolicyReconciler) Reconcile(eventCtx context.Context, req ctrl
...
+		gwLabels := gatewayLabels(ctx, r.Client(), gwKey)
+		owner := rlpKey.String()
+
+		// create/update ratelimit filters patch
+		// fetch already existing filters patch or create a new one
+		filtersPatckKey := client.ObjectKey{Namespace: gwKey.Namespace, Name: rlFiltersPatchName(gwKey.Name)}
```
maleck13, line 170: nit: should this be ```filtersPatchKey```

controllers/apim/ratelimitpolicy_controller.go
rahulanand16nov: Gateways are referenced in the `<namespace>/<name>` format but if the namespace is not mentioned, Istio will look into VirtualService's namespace.

apis/apim/v1alpha1/ratelimitpolicy_types.go
```diff
@@ -77,6 +77,9 @@ type NetworkingRef struct {
 
 // RateLimitPolicySpec defines the desired state of RateLimitPolicy
 type RateLimitPolicySpec struct {
+	//+listType=map
```
eguzki, line 80: nice way to leverage builtin validation support of unique list (not scalar) elements

eguzki: Where did you find doc to do that? I can only find https://book.kubebuilder.io/reference/markers/crd-processing.html
But it does not clearly state it as a validation tool, but a processing behavior configuration.

eguzki:
Have you considered following the same pattern as AuthorizationPolicy and EnvoyFilter objects to select the workload to apply the conf?

Instead of specific name reference
```yaml
spec:
  networkingRef:
    - type: VirtualService
      name: toystore
```

It would be label selector
```yaml
spec:
  selector:
    matchLabels:
      version: v1
```

Same  as AuthPolicy and EnvoyFilter, the selector decides where to apply the rate limit policy. The selector will match with workloads in the same namespace as the rate limit policy. IMO the implementation would be much simpler. You do not need to read the VirtualService, no need to add annotations to the VirtualService objects.

apis/apim/v1alpha1/ratelimitpolicy_types.go
rahulanand16nov: I used the one you mentioned. Also, saw similar usage somewhere in the Kubernetes source code and thought it worth our use case.

rahulanand16nov:
@eguzki I was relying on HTTPRoutes from VirtualService but in this implementation, I saw it doesn't have to be that way. I can totally see label selectors in the spec compared to networking ref.

eguzki:
I wonder if RLP should have references  to VirtualService (or GatewayAPI HttpRoutes) or gateways with label selectors. The RLP has a `name` field which is a direct reference to VirtualService field name. However, at the same time, it feels like the approach should be with references to gateways. Mainly because the HTTP filters and the limitador cluster has nothing to do with the istio virtual services.

rahulanand16nov:
@eguzki That's spot on! `ownerRLPs` was created to allow the creation of RLPs in a different namespace to that of gateway's. But if we move RLP to the gateway's namespace, we can use `ownerRefs` from Kubernetes and simplify implementation as you mentioned.

rahulanand16nov:
> I wonder if RLP should have references to VirtualService (or GatewayAPI HttpRoutes) or gateways with label selectors. The RLP has a `name` field which is a direct reference to VirtualService field name. However, at the same time, it feels like the approach should be with references to gateways. Mainly because the HTTP filters and the limitador cluster has nothing to do with the istio virtual services.

Worth discussing in the `Kuadrant Technical Discussion` call?

eguzki:
If RLP uses label selectors, maybe it is hard to have one HTTP filter envoy patch per gateway.

Let's say we have two RLP, one with label selector `app=a` and the other with label selector `app=b`. The controller would create two EnvoyFilter objects (to setup HTTP filters and limitador cluster) with the same label selectors. Let's say there is a gteway with both labels, then both EnvoyFilter objects would apply their configuration to this one gateway. The gateway would ending up having two PRE rate limit and two POST rate limit http filters. This is clearly unwanted.

maleck13:
> 

right so we need to think about the user experience there. That would mean a developer would need to create be able to create the resource in the same namespace as the gateway was deployed. 
Whereas the user experience that Gateway API seem to offer is that you add extensions to the HTTPRoute. These extensions have to be created in the same namespace as the HTTPRoute (they are locally referenced objects). 
Do we want to provide that kind of experience here? IE I create a VirtualService, Ingress, Route and I link my auth and rate limiting to my Ingress object. The context and responsibility of actually applying the the configuration lies with the control plane. For Gateway API that will ultimately be Istio for now it will likely have to be done by Kuadrant controller. For Ingress / route / VirtualService Kuadrant controller can play that role and provide the same type of experience that we want to offer via Gateway API

rahulanand16nov:
`RateLimitPolicy` name and the amount of information that we are gaining networking ref leans towards authpolicy-like experience.

maleck13:
So I am a bit confused. With the AuthConfig, it is created in the same namespace as the routing object. So are we suggesting that then if the developer wants to add rate limiting, they need to now create a resource in the gateway namespace? That feels very disjointed.

eguzki:
> With the AuthConfig, it is created in the same namespace as the routing object.

I guess you mean `AuthorizationPolicy`.

The authorizationpolic's selector will match with workloads in the namespace where it is created. Workloads are pods, so effectively we can assume that the authorization policy will be applied to the gateways (envoys) deployed in the same namespace whose labels match. 

The routing objects, as I understand, are Ingress (bare k8s), Route (OCP) or coming Gateway API's HttpRoute. These objects are created in several namespaces. Not necessarily in the same namespace where the ingress gateway runs. Another agent is the ingress controller, a pod listening for Route/Ingress/HTTPRoute objects and injecting the configuration in the ingress gateways deployed in the cluster. My point here is that, the authorization policy needs to be deployed in the same NS as the gateway, not the routing objects.

Please, correct me if I am wrong.

eguzki:
RateLimit configurations are linked to the Envoy's VirtualHost (or the route component of the VH). So, it does not make sense to link RateLimitPolicy to a gateway itself, but instead, it should be linked to the VirtualService / HTTPRoute. 

How RateLimitPolicy are attached to VirtualService / HTTPRoute can be discussed. Currently the policy has a list of VirtualServices direct references (namespace can be referenced as well). For now, I think it is good enough. We can enhance if needed in a second iteration.

maleck13:
@eguzki I agree. I think we will probably iterate towards using the same approach as we have for authconfig. Add an annotation to the VS/HTTPRoute/Ingress to link it to a RLP

rahulanand16nov:
If RLP becomes like AuthPolicy, annotation on VS is an addon that we can leverage.

controllers/apim/ratelimitpolicy_controller.go
```diff
@@ -136,100 +145,113 @@ func (r *RateLimitPolicyReconciler) Reconcile(eventCtx context.Context, req ctrl
 
 func (r *RateLimitPolicyReconciler) reconcileWithVirtualService(ctx context.Context, vs *istio.VirtualService, rlp *apimv1alpha1.RateLimitPolicy) error {
 	logger := r.Logger()
+	rlpKey := client.ObjectKeyFromObject(rlp)
 
 	_, ok := vs.Annotations[VSAnnotationRateLimitPolicy]
```
eguzki, line 150: why is the VS being annotated? What's the purpose?

controllers/apim/utils.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// rlFiltersPatchName returns the name of EnvoyFilter adding in rate-limit filters to a gateway.
+func rlFiltersPatchName(gatewayName string) string {
+	return gatewayName + "-ratelimit-filters"
+}
+
+// ratelimitsPatchName returns the name of EnvoyFilter adding in ratelimits to a gateway.
+func ratelimitsPatchName(rlpName, gatewayName string) string {
+	return fmt.Sprintf("ratelimits-of-%s-in-%s", rlpName, gatewayName)
```
eguzki, line 30: The envoy filter contains `Host` which belongs to the VS. So I wonder if the EnvoyFitler Name should be indexed by the VS as well.

controllers/apim/ratelimitpolicy_controller.go
```diff
@@ -136,100 +145,113 @@ func (r *RateLimitPolicyReconciler) Reconcile(eventCtx context.Context, req ctrl
...
+			filtersPatchOwnerList = strings.Split(ogOwnerRlps, ownerRlpSeparator)
+		}
 
-			vHostName := host + ":80" // Istio names virtual host in this format: host + port
-			routePatch := routeRateLimitsPatch(vHostName, route.Name, route.Actions, route.Stage)
+		// add the owner name if not present already
+		if !common.Contains(filtersPatchOwnerList, owner) {
+			filtersPatchOwnerList = append(filtersPatchOwnerList, owner)
+		}
+		finalOwnerVal := strings.Join(filtersPatchOwnerList, ownerRlpSeparator)
```
eguzki, line 198: Have you considered the option of having one annotation per owner in the EnvoyFilter object?

```ỳaml
metadata:
  annotations:
     kuadrant.io/ownerRateLimitPolicies-rlp1: true
     kuadrant.io/ownerRateLimitPolicies-rlp2: true
     kuadrant.io/ownerRateLimitPolicies-rlp3: true
```

The value of the annotations is meaningless, but it feels easier to:
* read for a human
* easier to manage from the code. The operator to check if the owner already exists is straightforward.

eguzki (changes requested):

controllers/apim/ratelimitpolicy_controller.go
rahulanand16nov: To create a reconcile request if it changes.

eguzki: But it is not implemented, is it?

If it is not, I would leave for another PR.

Anyway, if you want to reconcile EnvoyFilters if the VS changes, the annotation is not strictly needed. There is another way. You just need to add a watch of the VS kind with mapper that gets the list of RLF objects filtering by the `spec.networkingRef`.

controllers/apim/utils.go
```diff
@@ -0,0 +1,81 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// gatewayLabels fetches labels of an Istio gateway identified using the given ObjectKey.
+func gatewayLabels(ctx context.Context, client client.Client, gwKey client.ObjectKey) map[string]string {
+	gateway := &istio.Gateway{}
+	if err := client.Get(ctx, gwKey, gateway); err != nil {
+		return map[string]string{}
+	}
+	return gateway.Labels
```
eguzki, line 20: It is not the labels of the gateway object. It's about the labels of the workload pod (i.e. the actual gateway process). 

the `gateway.spec.selector` should be returned

rahulanand16nov: Fixed

controllers/apim/ratelimitpolicy_controller.go
rahulanand16nov: Fixed

controllers/apim/ratelimitpolicy_controller.go
rahulanand16nov: I thought about doing it that way but it felt like a little extra work that can be solved easily with an annotation but if we find it in the future interacting with users need in a negative way, can follow this route.

rahulanand16nov:
@eguzki I'll come back to left-out points once we remove the `networkingRef` because it changes the context behind them.

Approvals before I hit the merge?

eguzki (approved):

PR #86, opened 2022-02-14
Remove networkingRef from RateLimitPolicy

rahulanand16nov:
This PR:
- Removes the `networkingRef` from RateLimitPolicy API.
- Gives EnvoyFilter creation/deletion to networking objects (VirtualService/HTTPRoute). This is done through operation-specific annotation added by networking object controllers on RateLimitPolicy. Based on which annotation is used, the RLP controller will try to reconcile EnvoyFilters. This separates the concerns related to rate limit only to the RLP controller and networking object controllers just need to request the operations. Plus, for the HTTPRoute controller, we just need to use the `ratelimitpolicyPredicate` in the manager and we'll get the same functionality as VirtualService does!
- Removes the `ownerRefs` on AuthorizationPolicy by VirtualService controller because of cross-reference limitation. (Handling of VirtualService deletion is pending)

## Verification Steps
- `make local-setup`
- `k apply -f examples/toystore/toystore.yaml`
- `k apply -f examples/toystore/virtualService.yaml`
- `k annotate virtualservice/toystore kuadrant.io/auth-provider=kuadrant-authorization`
This will create the AuthPolicy using the information in the VirtualService
```
k get authorizationpolicy -A
NAMESPACE         NAME                                 AGE
kuadrant-system   on-kuadrant-gateway-using-toystore   12s
```
- `k apply -f config/samples/apim_v1alpha1_ratelimitpolicy.yaml`
- `k annotate virtualservice/toystore kuadrant.io/ratelimitpolicy=toystore`
This will create `RateLimit` and `EnvoyFilter`:
```
k get envoyfilter -A
NAMESPACE         NAME                                                    AGE
kuadrant-system   kuadrant-gateway-ratelimit-filters                      3s
kuadrant-system   ratelimits-on-kuadrant-gateway-using-default-toystore   3s
```
```
k get ratelimit -A
NAMESPACE   NAME                             AGE
default     ratelimitpolicy-sample-limit-1   102s
```


rahulanand16nov:
updated verification steps and it's working as expected for me.

config/samples/apim_v1alpha1_ratelimitpolicy.yaml
```diff
@@ -1,30 +1,44 @@
...
+                descriptor_value: "yes"
+  rateLimits:
+    - stage: PREAUTH
       actions:
         - generic_key:
-            descriptor_key: admin
+            descriptor_key: vhaction
             descriptor_value: "yes"
+
   limits:
```
maleck13, line 39: do you think with this sample it is worth adding a set of limits for the vhost level?

controllers/apim/ratelimitpolicy_controller.go
```diff
@@ -108,61 +105,128 @@ func (r *RateLimitPolicyReconciler) Reconcile(eventCtx context.Context, req ctrl
...
-			return ctrl.Result{}, nil
-		case apimv1alpha1.NetworkingRefTypeVS:
-			vsNamespacedName := client.ObjectKey{
-				Namespace: rlp.Namespace, // VirtualService lookup is limited to RLP's namespace
-				Name:      networkingRef.Name,
-			}
+	// Operation specific annotations must be removed if they were present
+	updateRequired := false
+	// check for delete operation for virtualservice
+	if vsName, present := rlp.Annotations[KuadrantDeleteVSAnnotation]; present {
```
maleck13, line 111: can you explain the use case for this annotation?  Around deletion of the VS, we can add a finalizer to the virtual service and then if it is deleted it wont be removed until we remove the finalizer. This will give us chance to clean up any associated Istio resources

maleck13: @eguzki perhaps you can help @rahulanand16nov with this ?

rahulanand16nov: `delete` operation is used even when the annotation is removed from the VS.

config/samples/apim_v1alpha1_ratelimitpolicy.yaml
rahulanand16nov: Sure, I can add one for the vhost level as well.

controllers/apim/ratelimitpolicy_controller.go
```diff
@@ -108,61 +105,128 @@ func (r *RateLimitPolicyReconciler) Reconcile(eventCtx context.Context, req ctrl
...
+		if err := r.Client().Get(ctx, vsNamespacedName, &vs); err != nil {
+			if apierrors.IsNotFound(err) {
+				logger.Info("no VirtualService found", "lookup name", vsNamespacedName.String())
+				return ctrl.Result{}, nil
 			}
+			logger.Error(err, "failed to get VirutalService")
+			return ctrl.Result{}, err
+		}
+
+		if err := r.detachFromNetwork(ctx, vs.Spec.Gateways, vsKey, &rlp); err != nil {
```
maleck13, line 128: I think with the annotation based attachment, we would need a VS controller that watched VS. When the VS is deleted, it would call into a ratelimiting package to detach and an auth package. If the rl annotation is removed, then the VS controller reconciler would be called and it would again call to the ratelimiting package to detach. @eguzki @rahulanand16nov WDYT?

rahulanand16nov: Yup, it was part of the consideration and we can totally do that. The only opposite point was that the RLP controller does nothing but create limitador's `RateLimit` resources. The reasoning behind it was to separate the concerns related to ratelimit to the RLP controller only. If you and @eguzki think it's the better way to go then I'll move the code.

rahulanand16nov: Plus, I was thinking RLP can store in its status block about various networking object that particular RLP is "attached" to.

eguzki:
answering to https://github.com/Kuadrant/kuadrant-controller/pull/86#discussion_r807667693 cc @maleck13 

It is indeed a non-standard (in controllers patterns) way of notifying an event that is not being watched by the controller

For me it is ok for now, taking into consideration that we want to have something working ASAP. I think the implementation is "weak". But we both agree that is it a good compromise for having something working now and full fledged (not missing any use case as we expect now)

The implementation is not optimal, indeed. Mainly because Custom Resources represent states and not actions as we are adding in the annotations. But we cannot come up with something easier to implement. The complexity comes from the fact that the bare existence of the reconciled resources (EnvoyFitlers and Limitador RateLimits) depends on the existence not only of the RateLimitPolicy, but a reference from a VS (or HTTPRoute) to the RLP as well. This relationship model, which is UX consistent with gateway api extensions refs, adds extra complexity for the controllers to manage the lifecycle of the reconciled resources (EnvoyFitlers and Limitador RateLimits). The controller needs (maybe not entirely at the first implementation) to handle use cases when:

*    The VS annotation with the RLP reference is deleted -> delete EnvoyFilters and Limitador RateLimits
*    The VS annotation with the RLP reference is updated -> delete old EnvoyFilters and Limitador RateLimits and create new ones
*    The RLP is deleted -> delete EnvoyFilters and Limitador RateLimits for all VS with reference to this object being deleted
*    [MISSING] The RLP is updated -> reconcile EnvoyFilters and Limitador RateLimits (This use case will be implemented in another PR

config/samples/apim_v1alpha1_ratelimitpolicy.yaml
rahulanand16nov: Added

controllers/apim/ratelimitpolicy_controller.go
eguzki: > I think with the annotation based attachment, we would need a VS controller that watched VS. When the VS is deleted, it would call into a ratelimiting package to detach and an auth package. If the rl annotation is removed, then the VS controller reconciler would be called and it would again call to the ratelimiting package to detach. @eguzki

But, we do have a VS controller listening to events of the VS resources. https://github.com/Kuadrant/kuadrant-controller/pull/86/files#diff-9657787be5bdaae1595dcc9da8bd48808e19334779f7deeab86dc912d5d65478L181 Is that what you were asking for? 

Some context about this:

The reconciled resources, EnvoyFilters and RateLimits, needs data from both the RLP and the VirtualService (Host and gateway(s) namespace and names). As I see, both approaches have sense, i.e., implementing the reconciliation job from the VS controller or the RLP controller. 

However, I think we need to distinguish between reconciling EnvoyFilters and RateLimits and catching events of VS deletion, VS annotation update/delete, RLP update and so on. Regarding the VS events, it seems reasonable to be done in the VS controllers, whereas regarding thef RLP events, it should be done by the RLP controller. The reconciliation of the EnvoyFilters and RateLimits could be implemented in both controllers.  And why not reconciling EnvoyFilters and RateLimits when catching events? Let me try to explain as I see:

In the operator-framework controller pattern, the reconciliation job is done in the reconciler object. Each controller has one reconciler main type (assigned with the `builder.For(TYPE)` method). Besides the main type, each controller can register [event handlers](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.10.0/pkg/handler/eventhandler.go#L42)  for other types as well. These event handlers are meant for (and I quote):
```
// EventHandler enqueues reconcile.Requests in response to events (e.g. Pod Create).  EventHandlers map an Event 
// for one object to trigger Reconciles for either the same object or different objects - e.g. if there is an    
// Event for object with type Foo (using source.KindSource) then reconcile one or more object(s) with type Bar.  
```

Thus, I feel that is not right to do the reconciliation job (creating, updating, deleting reconciled resources based on desired state) from within the event handlers. I think they were not designed for that. One clear sign of that is the error handling of the event handler. There is not. So, basically you cannot tell to retry.

controllers/apim/virtualservice_controller.go
```diff
@@ -177,30 +170,13 @@ func normalizeStringMatch(sm *networkingv1alpha3.StringMatch) string {
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *VirtualServiceReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	rlpMapper := &rateLimitPolicyMapper{
+		K8sClient: r.Client(),
+		Logger:    r.Logger(),
+	}
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&istionetworkingv1alpha3.VirtualService{}, builder.WithPredicates(virtualServicePredicate())).
+		For(&istionetworkingv1alpha3.VirtualService{}, builder.WithPredicates(routingPredicate(rlpMapper))).
```
eguzki, line 178: As an enhancement, instead of combining authorization and ratelimit logic into a single predicate, a cleaner approach would be to add multiple predicates (single responsability principle).

eguzki: it is good enough as it is now

eguzki (approved):

PR #93, opened 2022-02-17
fix per-route ratelimit option

rahulanand16nov:
Some things you only learn after making the mistake :')

### DIFF
--- a/controllers/apim/ratelimitpolicy_controller.go
+++ b/controllers/apim/ratelimitpolicy_controller.go
@@ -498,7 +498,9 @@ func virtualHostRateLimitsPatch(vHostName string, rateLimits []*apimv1alpha1.Rat
 		"value": map[string]interface{}{
 			"rate_limits": common.EnvoyFilterRatelimitsUnstructured(rateLimits),
 			"typed_per_filter_config": map[string]interface{}{
-				// Note the following name is different from what we have given to our pre/post auth ratelimit filters
+				// Note the following name is different from what we have given to our pre/post-auth
+				// ratelimit filters. It's because you refer to the type of filter and not the name field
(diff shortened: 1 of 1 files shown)


controllers/apim/ratelimitpolicy_controller.go
```diff
@@ -503,6 +497,13 @@ func virtualHostRateLimitsPatch(vHostName string, rateLimits []*apimv1alpha1.Rat
 		"operation": "MERGE",
 		"value": map[string]interface{}{
 			"rate_limits": common.EnvoyFilterRatelimitsUnstructured(rateLimits),
+			"typed_per_filter_config": map[string]interface{}{
```
eguzki, line 500: can you share doc about this. 

It is hard to me to get from https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ratelimit/v3/rate_limit.proto#enum-extensions-filters-http-ratelimit-v3-ratelimitperroute-vhratelimitsoptions

rahulanand16nov: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-virtualhost

rahulanand16nov: I saw this example to learn how to use that field: https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/ext_authz_filter.html?highlight=per_filter#per-route-configuration

eguzki: Thanks, could you update the example above of a serialized EF with the new adding?

eguzki (approved):

PR #94, opened 2022-02-17
fix ratelimit namespace and handle deletion of rl policy

eguzki:
* Limitador RateLimit CRs created in the namespace of the limitador service instead of the RateLimitPolicy namespace.

The limitador operator applies rate limit CR configuration on the limitador service running in the same namespace of the CR

* Delete RateLimit CRs when RateLimitPolicy is deleted

# Verification steps

1. `make local-setup`
2. Deploy toystore ratelimitpolicy in `default` namespace
`k apply -f examples/toystore`

The kuadrant controller should have created ratelimit objects in the same namespace as the limitador service, i.e. `kuadrant-system`

```
$ k get ratelimits -A
NAMESPACE         NAME                 AGE
kuadrant-system   default-toystore-1   7s
kuadrant-system   default-toystore-2   7s
```
3. Remove the rate limit policy
```
k delete ratelimitpolicy toystore
```
Check that ratelimit CRs are also deleted

```
$ k get ratelimits -A
No resources found.
```

### DIFF
--- a/controllers/apim/utils.go
+++ b/controllers/apim/utils.go
@@ -31,6 +31,11 @@ func ratelimitsPatchName(gwName string, networkKey client.ObjectKey) string {
 	return fmt.Sprintf("ratelimits-on-%s-using-%s-%s", gwName, networkKey.Namespace, networkKey.Name)
 }
 
+// limitadorRatelimitsName returns the name of Limitador RateLimit CR.
+func limitadorRatelimitsName(objKey client.ObjectKey, idx int) string {
+	return fmt.Sprintf("%s-%s-%d", objKey.Namespace, objKey.Name, idx)
+}
+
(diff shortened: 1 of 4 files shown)


controllers/apim/utils.go
```diff
@@ -31,6 +31,11 @@ func ratelimitsPatchName(gwName string, networkKey client.ObjectKey) string {
 	return fmt.Sprintf("ratelimits-on-%s-using-%s-%s", gwName, networkKey.Namespace, networkKey.Name)
 }
 
+// limitadorRatelimitsName returns the name of Limitador RateLimit CR.
+func limitadorRatelimitsName(objKey client.ObjectKey, idx int) string {
+	return fmt.Sprintf("%s-%s-%d", objKey.Namespace, objKey.Name, idx)
```
eguzki, line 36: Maybe the template should be `rlp-%s-%s-%d` ?

PR #95, opened 2022-02-22
ratelimitpolicy status has virtualservices and gateways

eguzki:
https://issues.redhat.com/browse/KUADRANT-39

# What
The RateLimitPolicy CRD now has a `.status` field with the list of  VirtualServices having reference to that ratelimitpolicy. The status also includes the gateways. 

This PR does not include the reconciliation of EnvoyFilters and RateLimits when the spec of the RateLimitPolicy changes

# Why
* Provide information about the current VS having reference to that ratelimitpolicy
* Use that information to reconcile changes (currently used for removing associated EnvoyFilters when the RLP is deleted)

# Verification Steps

```
make local-setup
```

Create RateLimitPolicy and one virtual Service (toystory)

```
k apply -f examples/toystore
```

Annotate the VS to reference the RLP

```
k annotate virtualservice toystore kuadrant.io/ratelimitpolicy=toystore
```

Check the status of the RLP

```yaml
$ k get ratelimitpolicy toystore -o jsonpath='{.status}' | yq e -P
virtualservices:
  - gateways:
      - name: kuadrant-gateway
        namespace: kuadrant-system
    name: toystore
```


pkg/istio/envoy_filters.go
```diff
@@ -1,4 +1,4 @@
-package common
```
eguzki, line 1: Moving package to have package `github.com/kuadrant/kuadrant-controller/pkg/common` as a dependency for package `github.com/kuadrant/kuadrant-controller/apis/apim/v1alpha1` (not the other way around as it was until now)

rahulanand16nov (approved):
Working as expected. Once #90 is merged, we can also do the same for HTTPRoute.

PR #98, opened 2022-02-24
increase duration and add postauth limit

rahulanand16nov:
This PR:
- Increases example RLP expiration time to 30 seconds
- Adds postauth limit for testing/demo.


eguzki (approved):

rahulanand16nov:
cc @maleck13 ratelimiting case covered fully.

You can follow the same commands: https://gist.github.com/rahulanand16nov/d86d8b91b267a8e3f0c420c0e2b87779

PR #100, opened 2022-03-01
fix envoyfilter for ratelimit action

eguzki:
## What

The EnvoyFilter to add rate limits actions configuration at the route level was adding a trailing dot `.`. 

## Why 

Istio does not add any trailing dot `.` for the route name. 

Istio adds a route for each match ([HTTPMatchRequest](https://istio.io/latest/docs/reference/config/networking/virtual-service/#HTTPMatchRequest). The route name (in envoy's virtualhost route) will `${routeName}.{matchName}`.

For a virtualservice like this without a route name (only match name):
```yaml
---
apiVersion: networking.istio.io/v1alpha3
kind: VirtualService
metadata:
  name: toystore
  labels:
    app: toystore
spec:
  hosts:
    - "*.toystore.com"
  gateways:
    - kuadrant-system/kuadrant-gateway
  http:
    - match:
        - name: get-toy
          uri:
            exact: /toy
          method:
            exact: GET
      route:
        - destination:
            host: toystore.default.svc.cluster.local
            port:
              number: 80
```

The ratelimitpolicy should be like this (`.get-toy`):
```yaml
---
kind: RateLimitPolicy
metadata:
  name: toystore
spec:
  routes:
    - name: ".get-toy"
      rateLimits:
        - stage: BOTH
          actions:
            - generic_key:
                descriptor_key: get-toy
                descriptor_value: "yes"
```

### DIFF
--- a/controllers/apim/ratelimitpolicy_controller.go
+++ b/controllers/apim/ratelimitpolicy_controller.go
@@ -480,7 +480,7 @@ func routeRateLimitsPatch(vHostName string, routeName string, rateLimits []*apim
 					Vhost: &networkingv1alpha3.EnvoyFilter_RouteConfigurationMatch_VirtualHostMatch{
 						Name: vHostName,
 						Route: &networkingv1alpha3.EnvoyFilter_RouteConfigurationMatch_RouteMatch{
-							Name: "." + routeName, // Istio adds '.' infront of names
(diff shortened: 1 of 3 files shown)


maleck13 (approved):
changes look good to me

PR #112, opened 2022-03-18
fetch operations from virtualservice

rahulanand16nov:
This PR:
- Makes `name` field in the `rule` object a fetching trigger for operation from VirtualSerivce.
- Makes `rule` object take array of `Operation`s.
- Implements detach part of signaling.

## Verification Steps
1. `make local-setup`
3. `k apply -f examples/toystore/virtualService.yaml`
4. `k apply -f config/samples/apim_v1alpha1_ratelimitpolicy.yaml`
5. `k annotate virtualservice/toystore kuadrant.io/ratelimitpolicy=toystore`
6. `istioctl dashboard envoy <kuadrant-gateway-pod>.kuadrant-system`

You should be able to find the following config in the `postauth.wasm` filter which is fetched from `VirtualService` and not present in `RateLimitPolicy`.

```json
{
          "operation": [
            {
              "paths": [
                "/admin/toy"
              ],
              "methods": [
                "POST"
              ],
              "hosts": [
                "*.toystore.com"
              ]
            },
            {
              "paths": [
                "/admin/toy"
              ],
              "methods": [
                "DELETE"
              ],
              "hosts": [
                "*.toystore.com"
              ]
            }
          ],
          "actions": [
            {
              "generic_key": {
                "descriptor_value": "yes",
                "descriptor_key": "admin"
              }
            }
          ]
        }
```


apis/apim/v1alpha1/ratelimitpolicy_types.go
```diff
@@ -103,24 +103,29 @@ type RateLimit struct {
...
 type Rule struct {
+	// Name supports regex for fetching operations from routing resources
+	// For VirtualService, if route name matches, all the match requests are
+	// converted to operations internally. But specific match request names
+	// are also supported.
 	Name string `json:"name,omitempty"`
 	// Operation specifies the operations of a request
 	// +optional
-	Operation *Operation `json:"operation,omitempty"`
+	Operation []*Operation `json:"operations,omitempty"`
```
eguzki, line 121: nit: strutct field should also be named `Operations`?

config/samples/apim_v1alpha1_ratelimitpolicy.yaml
```diff
@@ -5,30 +5,22 @@ metadata:
   name: toystore
 spec:
   rules:
-    - name: get-toy
-      operation:
-        paths: ["/toy"]
-        methods: ["GET"]
-        hosts: ["*.toystore.com"]
+    - operations:
```
eguzki, line 8: `name` has been deleted. 

There is a name in the next rules item. Do we still need a name?

controllers/apim/ratelimitpolicy_controller.go
```diff
@@ -496,6 +484,37 @@ func (r *RateLimitPolicyReconciler) reconcileLimits(ctx context.Context, rlp *ap
 	return nil
 }
 
+// fetchOperationsFromVS captures match rules from VirtualService and fill into RateLimitPolicy
+func fetchOperationsFromVS(vs *networkingv1alpha3.VirtualService, rlp *apimv1alpha1.RateLimitPolicySpec) {
```
eguzki, line 488: I do not get this. This method is building up a ratelimit policy out of the VS. But the reate limit policy is owned by the customer, so the controller is modifying it and it shouldn't.  I guess I am missing something

pkg/istio/wasm.go
```diff
@@ -6,7 +6,7 @@ import (
 
 // wasm-shim API structs
 type Rule struct {
-	Operation *apimv1alpha1.Operation         `json:"operation"`
+	Operation []*apimv1alpha1.Operation       `json:"operation"`
```
eguzki, line 9: nit: `Operations` ?

eguzki (changes requested):

config/samples/apim_v1alpha1_ratelimitpolicy.yaml
rahulanand16nov: name is optional which is used to fetch operations from the VirtualService.

controllers/apim/ratelimitpolicy_controller.go
rahulanand16nov: It won't update the user-owned `RateLimitPolicy` on the cluster but instead, only the struct created from it that is passed to the wasm-shim.

pkg/istio/wasm.go
rahulanand16nov: Done

apis/apim/v1alpha1/ratelimitpolicy_types.go
rahulanand16nov: Done

eguzki (approved):

PR #120, opened 2022-03-24
Update apim_v1alpha1_ratelimitpolicy.yaml

maleck13:
removing the host from the sample. I want to link to this example so don't want to show a host there currently. There maybe a place for it in the future but which hosts can be added there should be governed/restricted by the targeted resource


rahulanand16nov:
Can you please review #116 instead? I have removed hosts there as well but don't expect it to work since changes are required in the shim now.

rahulanand16nov:
#116 is merged, you can use refer the same :)

Issue #127, opened 2022-03-30
WASM Modules should be loaded via an Image


maleck13:
**What**

When providing a WASM module, we should load it via an image registry such as quay. This will allow us to fall in line with how the rest of the components are delivered and also used the same processes for building, testing , scanning etc

rahulanand16nov:
Not possible when inserting filters with `EnvoyFilter` but is with `WasmPlugin` which is in alpha right now.

maleck13:
@alexsnaps

maleck13:
https://istio.io/latest/docs/reference/config/proxy_extensions/wasm-plugin/

rahulanand16nov:
https://github.com/Kuadrant/kuadrant-controller/pull/152 completes the issue.

PR #152, opened 2022-05-17
Remove EnvoyFilter and use WasmPlugin

rahulanand16nov:
This PR works on #127 
- Bumps Go version from `1.16.x` to `1.17.x` which was required by different dependencies.
- Swaps `EnvoyFilter` with `WasmPlugin` resource.
  - Only one EF resource is left and is moved out of the controller. It can also be removed after https://github.com/Kuadrant/limitador/issues/53 is fixed.
- Wrappes `AuthPolicy` structs until: https://github.com/kubernetes-sigs/controller-tools/issues/667 is closed.

## Verification
Just try any one of the user guides with rate-limiting and it will work :pray: 

### Dev notes
The new istio extension requires ->  istio.io/api v0.0.0-20220413180505-1574de06b7bd -> github.com/go-logr/logr v1.2.0  and this logr conflicts with controller-runtime 0.10. Controller-runtime needs to be updated to 0.11. Both controller-runtime 0.11 and istio.io/api v0.0.0-20220413180505-1574de06b7bd -> requires golang 1.17


pkg/istio/wasm.go
```diff
@@ -67,12 +69,12 @@ func PluginPolicyFromRateLimitPolicy(rlp *apimv1alpha1.RateLimitPolicy, pluginSt
 	return pluginPolicy
 }
 
-func MergeMapStringPluginPolicy(modified *bool, existing *map[string]PluginPolicy, desired map[string]PluginPolicy) {
```
eguzki, line 70: why desired should be a pointer to a map?

pkg/istio/wasm_plugins.go
```diff
@@ -0,0 +1,148 @@
...
+	if !ok {
+		return false, fmt.Errorf("%T is not a *istioextensionv1alpha3.WasmPlugin", desiredObj)
+	}
+
+	// Deserialize config into PluginConfig struct
+	existingConfigJSON, err := existing.Spec.PluginConfig.MarshalJSON()
+	if err != nil {
+		return false, err
+	}
+	existingPluginConfig := &PluginConfig{}
```
eguzki, line 122: why marshal and then unmarshal?

utils/local-deployment/limitador-cluster.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: networking.istio.io/v1alpha3
+kind: EnvoyFilter
+metadata:
+  name: kuadrant-kuadrant-gwapi-gateway-wasm-ratelimits
```
eguzki, line 4: and labels? to which Gateway is being applied?

utils/local-deployment/deploy-kuadrant-deps.sh
```diff
@@ -28,6 +28,7 @@ kubectl apply -f utils/local-deployment/istio-manifests/autogenerated/Base/Base.
 kubectl apply -f utils/local-deployment/istio-manifests/autogenerated/Base/Pilot/Pilot.yaml
 kubectl apply -f utils/local-deployment/istio-manifests/autogenerated/Base/Pilot/IngressGateways/IngressGateways.yaml
 kubectl apply -n "${KUADRANT_NAMESPACE}" -f utils/local-deployment/istio-manifests/kuadrant-gateway.yaml
+kubectl apply -n "${KUADRANT_NAMESPACE}" -f utils/local-deployment/limitador-cluster.yaml
```
eguzki, line 31: why remove the EF related to the cluster? This cannot be done by the operator. Only the control plane (kuadrant-controller) knows about which gateways are being affected by the rate limit policies.

eguzki (changes requested):

controllers/apim/httproute_controller.go
```diff
@@ -164,7 +154,15 @@ func (r *HTTPRouteReconciler) reconcileAuthPolicy(ctx context.Context, logger lo
 				Name:      getAuthPolicyName(gwName, hr.Name, "custom"),
 				Namespace: gwNamespace,
 			},
-			Spec: authPolicySpec,
+			Spec: securityv1beta1.AuthorizationPolicy{
```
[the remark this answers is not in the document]
rahulanand16nov: As Istio went from `gogo/protobuf` to `golang/protobuf`, it removed the possibility of copying types, which is why I moved the spec definition inside the loop.

Ref: `Make all references to types use pointers; golang protobuf types cannot be copied` (https://github.com/istio/istio/pull/38055)

pkg/istio/wasm.go
rahulanand16nov: I believe the whole map shouldn't be copied, right?

utils/local-deployment/deploy-kuadrant-deps.sh
rahulanand16nov: We talked about this during RLP slides work. Consider this as a pre-requisite before Kuadrant can work properly and we should focus on fixing the issue in the Limitador.

utils/local-deployment/limitador-cluster.yaml
rahulanand16nov: We have two gateways so didn't care about narrowing it down to a single gateway.

rahulanand16nov:
> Can we close #69 ?

Closed

utils/local-deployment/deploy-kuadrant-deps.sh
eguzki: For the time being we can remove it and leave it as a pre-requisite. But I think it is a limitation. 

1.- It is error prone. Kuadrant could be configuring envoy with wasm module for rate limiting and would not work if the limitador service is not added to the envoy's clusters.

2.- Gateways have a lifecycle. They live and die. Having the kuadrant managing limitador's service configuration into envoy's conf, ensures that binding is working

3.- As a potential feature, kuadrant could provide a default limitador service. But that default service could be overridden by a custom limitador externally managed. And the location of the custom limitador would be ... in the  RateLimitPolicy?? in the kuadrant CR?  

So what would be the set-up procedure?
* Deploy istio, API gateway resources, ingress gateway
* Deploy kuadrant operator (limitador and authorino operator are also deployed as dependencies)
* Create the Kuadrant CR. This is when the operator deploys limitador service in the same namespace of the kuadrant CR. 
* Configure the gateway with the location of the limitador service. Who does this taks? Cluster operator? Kuadrant operator? If it is the kuadrant operator, how does the kuadrant operator know about the ingress gateway (or envoy) namespace? 

Again, I am not against this change for now. It is a simplification and I love making KISS things. But clearly it creates a empty gap that it is not clear to me how and how is fulfilled.

cc @maleck13

eguzki (changes requested):

utils/local-deployment/deploy-kuadrant-deps.sh
eguzki: @rahulanand16nov I also did not understand about "fixing the issue in the Limitador". This is about configuring limitador service in envoy, isn't it?

pkg/istio/wasm.go
rahulanand16nov: TIL

utils/local-deployment/deploy-kuadrant-deps.sh
rahulanand16nov: We have that EnvoyFilter in the first place due to: https://github.com/Kuadrant/limitador/issues/53

controllers/apim/httproute_controller.go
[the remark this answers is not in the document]
eguzki: Sorry, I miss something. If types use pointers, much better for having a common spec being copied to multiple authpolicies:
```go
mySpec := AuthPolicySpec{}
for idx := range gateways {
   authPolicy := AuthPolicy {
       Spec: mySpec,
   }
}

utils/local-deployment/deploy-kuadrant-deps.sh
eguzki: The limitador service is automatically added to the envoy cluster by istio. Ok. I assume the following. From the WASM module, when you need to open a connection, you need to use the cluster name as reference to define the connection destination. Right? I was assuming that because I guess it is envoy that opens the connection on behalf of the wasm module. If so, how does the WASM filter know about the cluster name? Does the kuadrant controller need to figure out something like `outbound|8081||limitador.rahul-test.svc.cluster.local` ?

controllers/apim/httproute_controller.go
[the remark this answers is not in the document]
rahulanand16nov: I guess the only thing you're missing here is that protobuf types cannot be copied but in the example you showed above, AuthPolicySpec is going to be copied for each time the loop runs.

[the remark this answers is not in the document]
eguzki: :+1:

eguzki (approved):
This is breaking our current configuration of envoy using EnvoyFilter.

Merging it but, as agreed, the envoyfilter with the configuration of the limitador service should be added back.

rahulanand16nov:
As soon as https://github.com/Kuadrant/limitador/pull/73 is merged, I'll make the appropriate change to this PR and merge it.

rahulanand16nov:
Since https://github.com/Kuadrant/limitador/pull/73 is closed. We can merge this PR for now and will re-add `EnvoyFilter` in the next PR. The current state only works in the development setup.

PR #163, opened 2022-06-07
add back envoyfilter to add limitador cluster patch

rahulanand16nov:
#152 Had removed the `EnvoyFilter` resource but due to the reason mentioned on that PR, we need to add it back so that it continues to work when installed using the operator.

PR changes can be verified by trying any one of the user guides.


controllers/apim/ratelimitpolicy_controller.go
```diff
@@ -261,10 +266,11 @@ func (r *RateLimitPolicyReconciler) cleanUpOrphanWASMPlugins(ctx context.Context
 
 	RateLimitStages := []apimv1alpha1.RateLimitStage{apimv1alpha1.RateLimitStagePREAUTH, apimv1alpha1.RateLimitStagePOSTAUTH}
 	for _, gwKey := range notReferencedGatewayKeys {
+		wasmPluginDeleted := false
 		for _, stage := range RateLimitStages {
 			wasmKey := kuadrantistioutils.WASMPluginKey(gwKey, stage)
 
-			wasmplugin := &istioextensionv1alpha3.WasmPlugin{}
```
eguzki, line 267: downgrade?

controllers/apim/ratelimitpolicy_controller.go
```diff
@@ -274,7 +280,22 @@ func (r *RateLimitPolicyReconciler) cleanUpOrphanWASMPlugins(ctx context.Context
...
+			if err != nil {
+				return err
+			}
+		}
+
+		// finalize pre-requisite of WasmPlugin i.e. EnvoyFilter adding the limitador cluster entry
+		if wasmPluginDeleted {
+			ef := &istionetworkingv1alpha3.EnvoyFilter{}
+			efKey := client.ObjectKey{Namespace: gwKey.Namespace, Name: kuadrantistioutils.LimitadorClusterEnvoyFilterName}
+			err := r.Client().Get(ctx, efKey, ef)
```
eguzki, line 293: why not delete and skip isNotFound errors?

rahulanand16nov: Good find :)

controllers/apim/ratelimitpolicy_controller.go
eguzki: then the name should be fixed.  You were planning to do that in another PR?

rahulanand16nov: By name here, I mean the name of the import i.e. `istioextensionv1alpha1`. It's already in this PR.

PR #165, opened 2022-06-09
make limitador cluster name and shim image url configurable through env

rahulanand16nov:
This makes limitador cluster name (defaults to `rate-limit-cluster`) and wasm-shim image (defaults to `quay.io/rahanand/wasm-shim:latest`) configurable through environment variables.

### DIFF
--- a/pkg/istio/wasm.go
+++ b/pkg/istio/wasm.go
@@ -4,10 +4,11 @@ import (
 	"reflect"
 
 	apimv1alpha1 "github.com/kuadrant/kuadrant-controller/apis/apim/v1alpha1"
+	"github.com/kuadrant/kuadrant-controller/pkg/common"
 	gatewayapiv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
 )
 
-const PatchedLimitadorClusterName = "rate-limit-cluster"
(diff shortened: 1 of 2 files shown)


alexsnaps (approved):

pkg/istio/wasm.go
```diff
@@ -4,10 +4,11 @@ import (
 	"reflect"
 
 	apimv1alpha1 "github.com/kuadrant/kuadrant-controller/apis/apim/v1alpha1"
+	"github.com/kuadrant/kuadrant-controller/pkg/common"
 	gatewayapiv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
 )
 
-const PatchedLimitadorClusterName = "rate-limit-cluster"
```
eguzki, line 10: this value does not need to be read from env var. 

In order to avoid conflicts, it should be generated from the limitador's (port, service, namespace) values tuple. 

The data to be read from the env var is the limitador's (port, service, namespace), hardcoded here

https://github.com/Kuadrant/kuadrant-controller/blob/eb85eee25874a3610747f102d77a5b70e0264652/pkg/common/common.go#L43

https://github.com/Kuadrant/kuadrant-controller/blob/eb85eee25874a3610747f102d77a5b70e0264652/pkg/common/common.go#L31

@didierofrivia is currently working on that

Issue #155, opened 2022-05-30
Implement the new design for AuthPolicy using Gateway API Policy attachment


maleck13:
**What**

Based on the design outlined in https://github.com/Kuadrant/kuadrant-controller/issues/151, implement auth policy:

https://hackmd.io/GQAOxDRdQTudSlsWdhtiUg

**Demo**

- After Kuadrant is installed, I am able to create an AuthPolicy that protects all routes for the host specified by my HTTPRoute
- I am able to modify this policy to protect just the /admin/* routes and allow /public/* to not be authenticated
- When I delete the authpolicy, my HTTPRoute should no longer be protected. 

- [x] Move proposal to this repo
- [x] Define the types for AuthPolicy and embed AuthConfig
- [x] Refactor the AuthPolicy controller
- [x] Add unit tests
- [x] Add e2e test 
- [x] Add documentation and basic guides for getting started with the AuthPolicy

maleck13:
@rahulanand16nov do you feel this is ready to implement?

rahulanand16nov:
Yeah, just need to sketch up the status part but otherwise looks ready to me.

rahulanand16nov:
Done :)

PR #166, opened 2022-06-10
implement new authpolicy design

rahulanand16nov:
#155 

-----
## Verification Steps:
1. `make local-setup`
2. `kubectl apply -f examples/toystore/toystore.yaml`
3. `kubectl apply -f examples/toystore/httproute.yaml`
4. Configure Authorino

```
kubectl apply -f -<<EOF
apiVersion: v1
kind: Secret
metadata:
  name: api-kek
  labels:
    authorino.kuadrant.io/managed-by: authorino
    group: friends
stringData:
  api_key: ITSFORDEMO
type: Opaque
EOF
```
5. `kubectl apply -f config/samples/apim_v1alpha1_authpolicy.yaml`
```
kubectl get authorizationpolicy -A
NAMESPACE         NAME                                       AGE
kuadrant-system   on-kuadrant-gwapi-gateway-using-toystore   9s

kubectl get authconfig -A
NAMESPACE         NAME                    READY
kuadrant-system   ap-default-toystore-1   true
```
7. `curl -v -XPOST -H 'Authorization: APIKEY ITSFORDEMO' -H 'Host: api.toystore.com' http://localhost:9080/admin/toy` // sweet 200 OK :)
8. `curl -v -XPOST -H 'Authorization: APIKEY WRONG' -H 'Host: api.toystore.com' http://localhost:9080/admin/toy` // 401 unauthorized
9. `kubectl delete -f config/samples/apim_v1alpha1_authpolicy.yaml`
```
kubectl get authorizationpolicy -n kuadrant-system
No resources found in kuadrant-system namespace.
```

## Next Step

- [ ] Target Gateway resource.
- [ ] Take hosts from GWAPI resources and limit its configurability.
- [ ] Status block for AuthPolicy


guicassolato (approved):
LGTM.

I wonder in the future whether Istio's `EnvoyFilter` resource won't be a better fit for us compared to (more limited) `AuthorizationPolicy`. (Consider, e.g., the use case for [host override via context extension](https://github.com/Kuadrant/authorino/blob/0c82786f2e93b5f4efd78c828326046143dd83f7/docs/user-guides/host-override.md).)

However, this is for now already an excellent start. Moreover, replacing (or not) Istio's AP in the implementation in the future is a detail that can be completely abstracted from the user, thanks to the simple and straightforward API offered here.

Good job, @rahulanand16nov!

rahulanand16nov:
> I wonder in the future whether Istio's `EnvoyFilter` resource won't be a better fit for us compared to (more limited) `AuthorizationPolicy`. (Consider, e.g., the use case for [host override via context extension](https://github.com/Kuadrant/authorino/blob/0c82786f2e93b5f4efd78c828326046143dd83f7/docs/user-guides/host-override.md).)

We had earlier implemented extAuthz API in the `wasm-shim` as well, but it was later removed. We can support all the cases with just one wasm module, i.e., pre-auth RL, auth (with context extension), and post-auth RL. Since context extension is per route thing, we come across the same problem of having an ID on the rules to match upon by `EnvoyFilter`, which to be is a bad UX.

PR #173, opened 2022-06-29
AuthPolicy: Add status block

rahulanand16nov:
#155 

This PR:
- Makes `AuthSchemes` singular so that only one `AuthScheme` is allowed hence `AuthConfig`.
- Adds status block to the `AuthPolicy` CRD.
- Updates `AuthPolicy` doc to be updated with the recent changes.

```
  status:
    conditions:
    - lastTransitionTime: "2022-06-29T12:23:14Z"
      message: HTTPRoute is protected
      reason: HTTPRouteProtected
      status: "True"
      type: Available
    observedGeneration: 1
```
```
  status:
    conditions:
    - lastTransitionTime: "2022-06-29T12:23:14Z"
      message: Gateway is protected
      reason: GatewayProtected
      status: "True"
      type: Available
    observedGeneration: 1
```

Note: If for some reason, `AuthConfig` is not ready when `AuthPolicy` is reconciling, it will say `AuthConfigNotReady`, but the status update will only happen in the next reconciliation event because IIRC, we cannot watch the status block of a resource (correct if I am wrong here).


rahulanand16nov:
> Are we consistent with the RLP? I am assuming yes but wanted to check that the user experience is at least similar from a status point of view for both

Yes, as of now.

eguzki (approved):
Looking good. 

Dropped some small comments

apis/apim/v1alpha1/common.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+// Interface to be implemented by Policy's status struct
+//+kubebuilder:object:generate=false
+type PolicyStatus interface {
+	GetObservedGeneration() int64
+	GetConditions() []metav1.Condition
+}
+
+func (r RateLimitPolicyStatus) GetObservedGeneration() int64 {
```
eguzki, line 17: this method in its own ratelimitpolicy_status.go file?

eguzki:
is README.md doc up to date regarding auth?

apis/apim/v1alpha1/common.go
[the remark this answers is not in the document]
rahulanand16nov: I think it can be either and only how you use it varies with it. If required in the future for a pointer receiver then no problem in changing it to that :)

controllers/apim/authpolicy_controller.go
```diff
@@ -83,45 +84,66 @@ func (r *AuthPolicyReconciler) Reconcile(eventCtx context.Context, req ctrl.Requ
...
+	return ctrl.Result{}, nil
+}
+
+func (r *AuthPolicyReconciler) reconcileSpec(ctx context.Context, ap *apimv1alpha1.AuthPolicy) (ctrl.Result, error) {
 	if err := ap.Validate(); err != nil {
 		return ctrl.Result{}, err
 	}
 
-	if err := r.reconcileNetworkResourceBackReference(ctx, &ap); err != nil {
+	if err := r.reconcileNetworkResourceBackReference(ctx, ap); err != nil {
```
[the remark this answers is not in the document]
rahulanand16nov: Are you referring to `ap` here? If yes, then `ap` is a pointer, so no copying happening here (except the pointer address), right?

apis/apim/v1alpha1/common.go
rahulanand16nov: Types are in `v1alpha1` package and implementing the interface in a different package is not allowed in Go but if you mean to create a new file called `ratelimitpolicy_status.go` then I'd say we can do that once `common.go` is large enough to break into different files.

rahulanand16nov:
> is README.md doc up to date regarding auth?

Docs update coming in next PR :)

controllers/apim/authpolicy_controller.go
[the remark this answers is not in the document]
eguzki: right, It is a pointer in the `reconcileSpec` method while in `Reconcile` is not. Hence my confusion.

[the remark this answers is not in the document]
rahulanand16nov: Yup, it's not a pointer in `Reconcile`, and it's fine to pass its address along to different functions. Maybe I am missing something about Golang? :sweat_smile:

apis/apim/v1alpha1/common.go
eguzki: Sorry, I was referring to the `ratelimitpolicy_types.go`.  So the implementation of the interface happens in its own file for rate lmit types and auth types.

>  implementing the interface in a different package is not allowed in Go 

What? where is that statement come from? I am missing something?

`myinterface/myinterface.go` file
```go
package myinterface

type PolicyStatus interface {
	GetObservedGeneration() int64
	GetConditions() []int64
}
```

`myinterfaceimpl/myinterfaceimpl.go` file

```go
package myinterfaceimpl

type PolicyStatusImpl int64

func (s PolicyStatusImpl) GetObservedGeneration() int64 {
	return int64(s)
}

func (s PolicyStatusImpl) GetConditions() []int64 {
	return []int64{int64(s)}
}
```

`main.go` file

```go
package main

import (
	"fmt"
	"my-module/myinterface"
	"my-module/myinterfaceimpl"
)

func main() {
	var xs myinterface.PolicyStatus = myinterfaceimpl.PolicyStatusImpl(5)
	fmt.Println(xs)
}
```

```
❯ go run main.go 
5
```

rahulanand16nov: I was previously trying to implement the interface in the `controllers/apim` package for status structs that are in `v1alpha` package.
```
invalid receiver type "github.com/kuadrant/kuadrant-controller/apis/apim/v1alpha1".RateLimitPolicyStatus (type not defined in this package)
```

eguzki: What you cannot do is to implement a method for type A in a package where A is not defined. If your type is defined in some package P, the type methods can be defined in multiple files, but all belonging to package P

PR #174, opened 2022-06-29
Improve rlp proposal

eguzki:


doc/proposals/rlp-target-gateway-resource.md
```diff
@@ -52,18 +54,58 @@ spec:
...
 
 **Note:** No `namespace`/`domain` defined. Kuadrant controller will figure out.
 **Note:** There is no `PREAUTH`, `POSTAUTH` stage defined. Ratelimiting filter should be placed after authorization filter to enable authenticated rate limiting. In the future, `stage` can be implemented.
 
-## Kuadrant-controller's behavior 
-Multiple RLP objects can have references to a given Gateway. Similarly, multiple RLP objects can have references to a given HTTPRoute. The kuadrant controller will iterate over the referenced gateways and aggregate all the rate limit policies that apply for each gateway, including RLP targeting HTTPRoutes. 
+## Kuadrant-controller's behavior
+
+Multiple RLP objects can have references to a single Gateway.
+Similarly, multiple RLP objects can have references to a single HTTPRoute.
```
rahulanand16nov, line 66: I thought we decided on only one RLP per GWAPI object, right? I remember it was part of the policy attachment doc as well.

doc/proposals/rlp-target-gateway-resource.md
```diff
@@ -104,35 +146,76 @@ Proposed new design:
...
-  - rules:
-    - paths: ["/admin/toy"]
-      hosts: ["cars.toystore.com"]
-      methods: ["POST", "DELETE"]
-    configurations:
-    - actions:
-      - generic_key:
-          descriptor_value: yes
-          descriptor_key: admin
+failureModeDeny: true
```
rahulanand16nov, line 149: nit: let's follow the same format: `failure_mode_deny`

doc/proposals/rlp-target-gateway-resource.md
```diff
@@ -104,35 +146,76 @@ Proposed new design:
...
+* [*major*] more aligned with RLS: [multiple descriptors](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-virtualhost-rate-limits) structured by "rate limit configurations" with matching rules
+
 ## WASM-SHIM
 
-Each rate limit policy host has a logical name as well as a set of hostnames to be applied it based on the incoming request’s host header.
+Each rate limit policy has a logical name and a set of hostnames to activate it based on the incoming request’s host header.
+
+The WASM-SHIM builds a tree based data structure holding the rate limit policies.
+The longest (sub)domain match is used to select the policy to be applied.
+Only one policy is being applied per invocation.
```
rahulanand16nov, line 178: I think we mentioned the usage of two policies for a single request, right?

doc/proposals/rlp-target-gateway-resource.md
```diff
@@ -104,35 +146,76 @@ Proposed new design:
...
+configurations:
+- actions:
+   - generic_key:
+        descriptor_key: admin
+        descriptor_value: "1"
+```
+
+Each configuration produces, at most, one descriptor. Depending on the incoming request, one configuration may or may not produce a rate limit descriptor.
+
+Each policy configuration has associated, optionally, a set of rules to match. Rules allow matching `hosts` and/or `methods` and/or `paths`. Matching occurs when at least one rule applies against the incoming request. If rules are not set, it is equivalent to matching all the requests.
```
rahulanand16nov, line 199: > If rules are not set, it is equivalent to matching all the requests.

A route-level policy with no rules should still be under the hostnames provided in the targeted `HTTPRoute` and for gateway-level policy, we need to use `*` as the hostname.

doc/proposals/rlp-target-gateway-resource.md
eguzki: Yes, but I think we can implement that. Nothing prevents us for doing that. 

Take these words with care because we find some unexpected behavior or over complication and we need to fall back to single RLP per HTTPRoute and Gateway constraint. Evil is in the details.

doc/proposals/rlp-target-gateway-resource.md
eguzki: :+1: 

the actual implementation is already like that https://github.com/Kuadrant/wasm-shim/pull/19/files#diff-f40c086c43dffda8b3394ef8a513eb53944192a302db38fec7b224035388c668R89

doc/proposals/rlp-target-gateway-resource.md
eguzki: two policies? in what context?

doc/proposals/rlp-target-gateway-resource.md
rahulanand16nov: Of course, we can implement but I don't feel comfortable opening that door right now. Opening for RLP suggests considering the same for AP as well then.

doc/proposals/rlp-target-gateway-resource.md
rahulanand16nov: > Request 1 (`api.toystore.com`) -> apply RLP A and RLP G
Two policies A and G are being applied, right?

doc/proposals/rlp-target-gateway-resource.md
eguzki: at the wasm filter level where are no route level and gateway level policies. There are rate limit configurations. With hostnames to scope them. 

`rules` provides flexibility to decide if one "configuration" (`rate_limit_policies[*].gatewat_actions[*].configurations`) needs to be executed or not. But always for a given hostname. `hostnames` is like a configuration rule, but applied globally for all configurations at wasm filter level.

Rules allow to decide not only based on the `hosts`, but also `method` and `path`.

For example.

There is  a rate limit configuration for `*.toystore.com` and I want some rate limit descriptors configuration only for `api.toystore.com` and another set of descriptors for `admin.toystore.com`. The wasm filter config would look like this:

```yaml
failure_mode_deny: true
rate_limit_policies:
  - name: toystore
    rate_limit_domain: toystore-app
    upstream_cluster: rate-limit-cluster
    hostnames: ["*.toystore.com"]
    gateway_actions:
      - configurations:  # no rules. Applies to all *.toystore.com traffic
          - actions:
              - generic_key:
                  descriptor_key: toystore-app
                  descriptor_value: "1"
      - rules:
          - hosts: ["api.toystore.com"]
        configurations:
          - actions:
              - generic_key:
                  descriptor_key: api
                  descriptor_value: "1"
      - rules:
          - hosts: ["admin.toystore.com"]
        configurations:
          - actions:
              - generic_key:
                  descriptor_key: admin
                  descriptor_value: "1"
```

eguzki:
@rahulanand16nov I have added some content. Hopefully it is now more clear.

doc/proposals/rlp-target-gateway-resource.md
rahulanand16nov: Ahhh, make sense now. :+1:

doc/proposals/rlp-target-gateway-resource.md
rahulanand16nov: got it. I now see how it will work out :)

rahulanand16nov (approved):

PR #176, opened 2022-07-01
Improve rlp proposal

eguzki:
More aligned RateLimitPolicy structure with the WASM plugin configuration object

### DIFF
--- a/doc/proposals/rlp-target-gateway-resource.md
+++ b/doc/proposals/rlp-target-gateway-resource.md
@@ -26,7 +26,13 @@ The goal of this document is to define:
 
 ## Envoy's Rate Limit Service Potocol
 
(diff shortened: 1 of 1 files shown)


doc/proposals/rlp-target-gateway-resource.md
```diff
@@ -42,22 +42,39 @@ spec:
...
+    - rules:
         - paths: ["/admin/*"]
           methods: ["GET"]
           hosts: ["example.com"]
-      actions:
-        - generic_key:
-            descriptor_key: admin
-            descriptor_value: "yes"
+      configurations:
+        - actions:
```
rahulanand16nov, line 50: I am wondering what's the advantage for a user to turn this into a `list of actions` instead of just `actions`?

eguzki: To be more aligned with [config.route.v3.RouteAction](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-msg-config-route-v3-routeaction)

It allows you to define more than one descriptor. Not only one. And why that? Because that is more flexible way to define your rate limit. You may want to define one rate limit descriptor for each use case. For example, you may want to define a rate limit descriptor generic and another descriptor depending on some header. If the header does not exist, the second descriptor is not generated, but you still have the other descriptor to rate limit.

rahulanand16nov: I see, this is quite helpful then I think having this explanation would help out the user to write better RLPs :)

rahulanand16nov (approved):
LGTM

PR #184, opened 2022-07-14
update metadata action struct

rahulanand16nov:
Related to Kuadrant/wasm-shim#10


eguzki:
In order to close https://github.com/Kuadrant/wasm-shim/issues/10, user guide to include authenticated rate limit is needed

rahulanand16nov:
@eguzki updated authenticated ratelimit guide

apis/apim/v1alpha1/ratelimitpolicy_types.go
```diff
@@ -39,10 +39,14 @@ type GenericKeySpec struct {
...
 
-type MetadataPathSegment struct {
+type MetadataPathSegmentKey struct {
 	Key string `json:"key"`
 }
 
+type MetadataPathSegment struct {
+	Segment MetadataPathSegmentKey `json:"segment"`
+}
+
```
didierofrivia, line 49:
```suggestion
type MetadataPathSegmentKey string `json:"key"`

type MetadataPathSegment struct {
        Key MetadataPathSegmentKey `json:"key"`
}

```

or simply:

```suggestion
type MetadataPathSegment struct {
        Key string `json:"key"`
}

```

rahulanand16nov: We must follow the exact `structs` definition as in `wasm-shim`. That simplification might not produce the correct result required, I guess.

eguzki: According to [type.metadata.v3.MetadataKey](https://www.envoyproxy.io/docs/envoy/latest/api-v3/type/metadata/v3/metadata.proto#envoy-v3-api-msg-type-metadata-v3-metadatakey) it is 
```yaml
key: envoy.xxx
path:
- key: prop
- key: foo
```

So maybe we can get rid of that `segment`

rahulanand16nov:
```rust
pub struct MetadataKey {
    // message fields
    pub key: ::std::string::String,
    pub path: ::protobuf::RepeatedField<MetadataKey_PathSegment>,
    // special fields
    #[cfg_attr(feature = "with-serde", serde(skip))]
    pub unknown_fields: ::protobuf::UnknownFields,
    #[cfg_attr(feature = "with-serde", serde(skip))]
    pub cached_size: ::protobuf::CachedSize,
}


pub struct MetadataKey_PathSegment {
    // message oneof groups
    pub segment: ::std::option::Option<MetadataKey_PathSegment_oneof_segment>,
    // special fields
    #[cfg_attr(feature = "with-serde", serde(skip))]
    pub unknown_fields: ::protobuf::UnknownFields,
    #[cfg_attr(feature = "with-serde", serde(skip))]
    pub cached_size: ::protobuf::CachedSize,
}

pub enum MetadataKey_PathSegment_oneof_segment {
    key(::std::string::String),
}
```

rahulanand16nov: @eguzki Does that help answer the question?

eguzki: I am afraid, the `segment` is necessary. I do not understand why, though. 

If you try to parse (using the rust parser):
```json
"metadata": {                                      
    "descriptor_key": "user-id",                   
    "default_value": "no-user",                    
   "metadata_key": {                              
       "key": "envoy.filters.http.ext_authz",     
       "path": [                                  
           {                                      
               "key": "ext_auth_data"             
           }, 
           {                                      
               "key": "user_id"             
           }                                    
       ]                                          
    },                                             
    "source": "DYNAMIC"         
}
```

The parser does not return error. And it should IMO. However, when you read the data
```rust
assert_eq!(metadata_path.len(), 2);                     // OK
assert_eq!(metadata_path[0].get_key(), "ext_auth_data");   // NOT OK
```
you get
```
thread 'configuration::test::parse_config' panicked at 'assertion failed: `(left == right)`
  left: `""`,
 right: `"ext_auth_data"`', src/configuration.rs:185:13
```

rahulanand16nov: Yup, that's the reasoning behind that test as well. I saw it when I was running limitador in debug mode and keys aren't present if the segment is not provided.

rahulanand16nov: We can, if we want to, remove `segment` part for the users by custom serialization from manifest to wasm config.

apis/apim/v1alpha1/ratelimitpolicy_types.go
```diff
@@ -53,8 +57,8 @@ type MetadataSpec struct {
 	MetadataKey   MetadataKeySpec `json:"metadata_key"`
 	// +optional
 	DefaultValue *string `json:"default_value,omitempty"`
-	// +optional
```
eguzki, line 56: It is still optional, right? I think that kubebuilder does not do anything with `// +optional` comment, but I would keep it as for the reader it makes it very clear.

The `// +kubebuilder:default=DYNAMIC` comment only affects deserialization, right? on serialization, if it is empty string `""`) it will just not be added in the json strinc.

rahulanand16nov: I think default itself means optional because if it's not provided in the manifest, it will be defaulted to `DYNAMIC`.

controllers/apim/ratelimitpolicy_controller.go
```diff
@@ -49,6 +49,9 @@ type RateLimitPolicyReconciler struct {
 //+kubebuilder:rbac:groups=apim.kuadrant.io,resources=ratelimitpolicies,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=apim.kuadrant.io,resources=ratelimitpolicies/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=apim.kuadrant.io,resources=ratelimitpolicies/finalizers,verbs=update
+//+kubebuilder:rbac:groups=limitador.kuadrant.io,resources=limitadors,verbs=get;list;watch;create;update;patch;delete
```
eguzki, line 52: :+1

eguzki (approved):

PR #195, opened 2022-08-29
RLP controller: Fix gateway not found

eguzki:
### what
the RLP controller handles when a referenced gateway does not exist. However, the finalizer part of the controller (the code that handles the cleaning up of resources when the RLP is deleted) did not handle correctly when a gateway did not exist. This PR, fixes that. Without the fix, the controller could not remove the finalizer in the CR and the deletion of the CR was blocked. 

### verification steps
Deploy the kuadrant controller

```
make local-setup
```

Create RLP with broken gateway reference

```yaml
k apply -f - <<EOF
---
kind: RateLimitPolicy
metadata:
  name: global-settings
spec:
  targetRef:
    kind: Gateway
    name: not-existing-name
  rateLimits:
    - configurations:
        - actions:
            - generic_key:
                descriptor_key: global-traffic
                descriptor_value: "1"
      limits:
        - conditions:
            - "global-traffic == 1"
          maxValue: 100
          seconds: 10
          variables: []
EOF
```
Delete the rlp. The deletion should not be blocked

```
k delete ratelimitpolicy global-settings
```

### DIFF
--- a/apis/apim/v1alpha1/authpolicy_types.go
+++ b/apis/apim/v1alpha1/authpolicy_types.go
@@ -40,8 +40,8 @@ type AuthPolicyStatus struct {
 	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,1,rep,name=conditions"`
 }
 
-//+kubebuilder:object:root=true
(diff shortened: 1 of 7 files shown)


didierofrivia (approved):
🐛 🔫

PR #198, opened 2022-09-02
support gateways RLPs alone

### LINKED ISSUE #197: Allow RLP targeting gateway without RLP targeting a route
Currently rate limit policies targeting a gateway are "merged" with policies targeting HTTPRoutes for that gateway. 

If there is no RLP targeting an HTTPRoute for a given gateway, the RLP targeting that gateway is not being applied

Kuadrant must enforce rate limit policies targeting a gateway even when there are no RLPs targeting HTTPRoutes. The existence of an HTTPRoute is enough to apply the gateway rate limit policies to the route's workload.

eguzki:
### what

Fix #197 

### how

Before merging route and gateway level limits/{gateway actions}, add gateway level limits/{gateway actions} for the hostnames defined in the gateway (falling back to global wildcard `*` if not defined)

### verification steps

Deploy the kuadrant controller

```
make local-setup
```
* deploy toystore service
```
kubectl apply -f examples/toystore/toystore.yaml
```
Create `toystore` HTTPRoute to configure routing to the toystore service

```yaml
kubectl apply -f - <<EOF
---
kind: HTTPRoute
metadata:
  name: toystore
  labels:
    app: toystore
spec:
  parentRefs:
    - name: istio-ingressgateway
      namespace: istio-system
  hostnames: ["*.toystore.com"]
  rules:
    - matches:
        - path:
            type: PathPrefix
            value: "/toy"
          method: GET
        - path:
            type: PathPrefix
            value: "/free"
          method: GET
        - path:
            type: Exact
            value: "/admin/toy"
          method: POST
      backendRefs:
        - name: toystore
          port: 80
EOF
```

Check `toystore` HTTPRoute works and it is not rate limited.

`GET /toy`: no rate limiting
```
while :; do curl --write-out '%{http_code}' --silent --output /dev/null -H "Host: api.toystore.com" http://localhost:9080/toy | egrep --color "\b(429)\b|$"; sleep 1; done
```

`POST /admin/toy`: no rate limiting
```
while :; do curl --write-out '%{http_code}' --silent --output /dev/null -H "Host: api.toystore.com" -X POST http://localhost:9080/admin/toy | egrep --color "\b(429)\b|$"; sleep 1; done
```

Adding gateway level RLP

| Policy | Rate Limits |
| ------------- | -----: |
| `POST /*` | **2** reqs / **10** secs (0.2 rps) |**

```yaml
kubectl apply -f - <<EOF
---
kind: RateLimitPolicy
metadata:
  name: istio-gw
  namespace: istio-system
spec:
  targetRef:
    kind: Gateway
    name: istio-ingressgateway
  rateLimits:
    - rules:
      - methods: ["POST"]
      configurations:
        - actions:
            - generic_key:
                descriptor_key: expensive_op
                descriptor_value: "1"
      limits:
        - conditions: ["expensive_op == '1'"]
          maxValue: 2
          seconds: 10
          variables: []
EOF
```

Validating the gateway level rate limit policy:

`GET /toy` : no rate limiting 
```
while :; do curl --write-out '%{http_code}' --silent --output /dev/null -H "Host: api.toystore.com" http://localhost:9080/toy | egrep --color "\b(429)\b|$"; sleep 1; done
```

`POST /admin/toy`:  expected to be rate limited @ **2** reqs / **10** secs (0.2 rps)
```
while :; do curl --write-out '%{http_code}' --silent --output /dev/null -H "Host: api.toystore.com" -X POST http://localhost:9080/admin/toy | egrep --color "\b(429)\b|$"; sleep 1; done
```

### DIFF
--- a/controllers/apim/ratelimitpolicy_cluster_envoy_filter.go
+++ b/controllers/apim/ratelimitpolicy_cluster_envoy_filter.go
@@ -70,21 +70,6 @@ func (r *RateLimitPolicyReconciler) gatewayRateLimitingClusterEnvoyFilter(
 	gwKey := client.ObjectKeyFromObject(gw)
 	logger.V(1).Info("gatewayRateLimitingClusterEnvoyFilter", "gwKey", gwKey, "rlpRefs", rlpRefs)
 
-	// Load all relevant rate limit policies
-	routeRLPList := make([]*apimv1alpha1.RateLimitPolicy, 0)
-	for _, rlpKey := range rlpRefs {
-		rlp := &apimv1alpha1.RateLimitPolicy{}
-		err := r.Client().Get(ctx, rlpKey, rlp)
-		logger.V(1).Info("gatewayRateLimitingClusterEnvoyFilter", "get rlp", rlpKey, "err", err)
-		if err != nil {
-			return nil, err
-		}
-
-		if rlp.IsForHTTPRoute() {
-			routeRLPList = append(routeRLPList, rlp)
-		}
-	}
-
 	ef := &istioclientnetworkingv1alpha3.EnvoyFilter{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "EnvoyFilter",
@@ -102,7 +87,7 @@ func (r *RateLimitPolicyReconciler) gatewayRateLimitingClusterEnvoyFilter(
 		},
 	}
 
-	if len(routeRLPList) < 1 {
+	if len(rlpRefs) < 1 {
 		common.TagObjectToDelete(ef)
 		return ef, nil
 	}
--- a/controllers/apim/ratelimitpolicy_finalizers.go
+++ b/controllers/apim/ratelimitpolicy_finalizers.go
@@ -65,10 +65,7 @@ func (r *RateLimitPolicyReconciler) computeFinalizeGatewayDiff(ctx context.Conte
 		LeftGateways: nil,
 	}
 
-	var rlpGwKeys []client.ObjectKey
-	var err error
-
-	rlpGwKeys, err = r.rlpGatewayKeys(ctx, rlp)
+	rlpGwKeys, err := r.rlpGatewayKeys(ctx, rlp)
 	if err != nil && !apierrors.IsNotFound(err) {
 		return nil, err
 	}
--- a/controllers/apim/ratelimitpolicy_finalizers.go
+++ b/controllers/apim/ratelimitpolicy_finalizers.go
@@ -43,7 +43,7 @@ func (r *RateLimitPolicyReconciler) finalizeRLP(ctx context.Context, rlp *apimv1
 		return err
 	}
 
-	if err := r.reconcileLimits(ctx, rlp, gatewayDiffObj); err != nil {
+	if err := r.reconcileLimits(ctx, rlp, gatewayDiffObj); err != nil && !apierrors.IsNotFound(err) {
 		return err
 	}
 
@@ -65,8 +65,11 @@ func (r *RateLimitPolicyReconciler) computeFinalizeGatewayDiff(ctx context.Conte
 		LeftGateways: nil,
 	}
 
-	rlpGwKeys, err := r.rlpGatewayKeys(ctx, rlp)
-	if err != nil {
+	var rlpGwKeys []client.ObjectKey
+	var err error
+
+	rlpGwKeys, err = r.rlpGatewayKeys(ctx, rlp)
+	if err != nil && !apierrors.IsNotFound(err) {
 		return nil, err
 	}
 
--- a/controllers/apim/ratelimitpolicy_limits.go
+++ b/controllers/apim/ratelimitpolicy_limits.go
@@ -3,6 +3,7 @@ package apim
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 
 	"github.com/go-logr/logr"
 	limitadorv1alpha1 "github.com/kuadrant/limitador-operator/api/v1alpha1"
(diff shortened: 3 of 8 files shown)


controllers/apim/ratelimitpolicy_cluster_envoy_filter.go
```diff
@@ -70,21 +70,6 @@ func (r *RateLimitPolicyReconciler) gatewayRateLimitingClusterEnvoyFilter(
 	gwKey := client.ObjectKeyFromObject(gw)
 	logger.V(1).Info("gatewayRateLimitingClusterEnvoyFilter", "gwKey", gwKey, "rlpRefs", rlpRefs)
 
-	// Load all relevant rate limit policies
```
eguzki, line 73: this changes has nothing to do with the purpose of the PR. This code was just useless.

controllers/apim/ratelimitpolicy_finalizers.go
```diff
@@ -65,8 +65,11 @@ func (r *RateLimitPolicyReconciler) computeFinalizeGatewayDiff(ctx context.Conte
 		LeftGateways: nil,
 	}
 
-	rlpGwKeys, err := r.rlpGatewayKeys(ctx, rlp)
-	if err != nil {
+	var rlpGwKeys []client.ObjectKey
+	var err error
```
didierofrivia, line 69: I see this change quite explicit here, but I'm failing to see why it's needed to define them instead of the way it was

controllers/apim/ratelimitpolicy_limits.go
```diff
@@ -115,23 +118,35 @@ func (r *RateLimitPolicyReconciler) gatewayLimits(ctx context.Context, rlpRefs [
 		if rlp.IsForHTTPRoute() {
 			routeRLPList = append(routeRLPList, rlp)
 		} else if rlp.IsForGateway() {
-			gwRLPList = append(gwRLPList, rlp)
+			if gwRLP == nil {
+				gwRLP = rlp
```
didierofrivia, line 122: Is there any difference in the spec between GW RLP and Route RLP?

controllers/apim/ratelimitpolicy_limits.go
```diff
@@ -115,23 +118,35 @@ func (r *RateLimitPolicyReconciler) gatewayLimits(ctx context.Context, rlpRefs [
...
 
 	limits := rlptools.LimitsByDomain{}
 
-	// iterate over HTTPRoute RLP's.
-	// Gateway level RLP's alone do not make sense.
-	// The request would be rejected by the router with 404
+	if gwRLP != nil {
+		if len(gw.Hostnames()) == 0 {
+			// wildcard domain
+			limits["*"] = append(limits["*"], gwRLP.FlattenLimits()...)
```
didierofrivia, line 134: I reckon this is because the Route RLP was copied to the GW RLP, but the former doesn't have in the Spec "Hostnames", so we apply all the limits to the arbitrary wildcard domain.. would that be correct?

didierofrivia: I'd be tempted to extract this logic in to a fn, which name would be descriptive of what it does

controllers/apim/ratelimitpolicy_finalizers.go
eguzki: me neither. 

I will fix that

controllers/apim/ratelimitpolicy_limits.go
eguzki: No, the spec is the same. It is only the target. Some RLP target a HTTPRoute and others a Gateway.

The list of RLP's come from some annotations, so here the code just separate them looking at the target.

controllers/apim/ratelimitpolicy_limits.go
eguzki: we need to add gateway limits to the index which is indexed by the domain. If the gateway does not specify any domain, global wilcard domain is assumed. and the code creates a new entry with the gateway limits indexed by the wilcard domain `*`

didierofrivia: I see, makes sense, another way would be to have a validating webhook returning the error that the hostname is missing

controllers/apim/ratelimitpolicy_wasm_plugins.go
```diff
@@ -116,45 +120,52 @@ func (r *RateLimitPolicyReconciler) gatewayWASMPlugin(ctx context.Context, gw *g
...
 
 		if rlp.IsForHTTPRoute() {
 			routeRLPList = append(routeRLPList, rlp)
 		} else if rlp.IsForGateway() {
-			gwRLPList = append(gwRLPList, rlp)
+			if gwRLP == nil {
+				gwRLP = rlp
+			} else {
+				return nil, fmt.Errorf("wasmPluginConfig: multiple gateway RLP found and only one expected. rlp keys: %v", rlpRefs)
+			}
```
didierofrivia, line 144: Maybe in a future refactor we could extract this sorting of the RLPs in a reusable fn, `sortRLPs` or something similar, to use also in `gatewayLimits`

controllers/apim/ratelimitpolicy_wasm_plugins.go
```diff
@@ -116,45 +120,52 @@ func (r *RateLimitPolicyReconciler) gatewayWASMPlugin(ctx context.Context, gw *g
...
+		if len(gw.Hostnames()) == 0 {
+			// wildcard domain
+			gatewayActions["*"] = append(gatewayActions["*"], rlptools.GatewayActionsFromRateLimitPolicy(gwRLP)...)
+		} else {
+			for _, gwHostname := range gw.Hostnames() {
+				gatewayActions[gwHostname] = append(gatewayActions[gwHostname], rlptools.GatewayActionsFromRateLimitPolicy(gwRLP)...)
+			}
+		}
+	}
+
```
didierofrivia, line 160: This is also a candidate for future refactor, the defaulting to `*` or having a validating webhook could be reuse here too

didierofrivia (approved):
Among the comments left and discussing the possible changes in the near future, this one is good to go 👍🏼

controllers/apim/ratelimitpolicy_limits.go
eguzki: well, if the hostname is missing, there is nothing wrong with it. It is just equivalent to `*`

Issue #190, opened 2022-08-18
RateLimitPolicy and AuthPolicy rules proposal


maleck13:
**What**

within RLP and AP, we have a rules section. I want to propose that we allow these to be empty and take the default values from the target resource. For AuthPolicy that would also include the hosts for the AuthConfig.

In RateLimitPolicy this would mean, the below would be applied to any host and path / method combination specified in the target HTTPRoute. When targeting a gateway it would apply to any host the gateway could handle. We can still allow the user to be more specific if they need to be, but this feels like a good default

```
apiVersion: apim.kuadrant.io/v1alpha1
kind: RateLimitPolicy
metadata:
  name: toystore-httproute
spec:
  targetRef:
    group: gateway.networking.k8s.io
    kind: HTTPRoute
    name: toystore
  rateLimits:
    - rules:
      configurations:
        - actions:
            - generic_key:
                descriptor_key: get-toy
                descriptor_value: "yes"
      limits:
        - conditions: ["get-toy == yes"]
          maxValue: 2
          seconds: 30
          variables: []

```

AuthPolicy could be created like this targeting a HTTPRoute with "api.toystore.com and admin.toystore" as its host but instead of repeating that information look like the below

```
piVersion: apim.kuadrant.io/v1alpha1
kind: AuthPolicy
metadata:
  name: toystore
  namespace: kuadrant-system
spec:
  targetRef:
    group: gateway.networking.k8s.io
    kind: Gateway
    name: kuadrant-gwapi-gateway
  authScheme:
    identity:
      - name: apikey
        apiKey:
          labelSelectors:
            app: toystore
            admin: "yes"
        credentials:
          in: authorization_header
          keySelector: APIKEY
```

Then, if the end user needs to be more specific (IE only target admin.toystore.com) they can fill in the rules. Perhaps when taking the values from the target, we add that as detail to the status?

maleck13:
@rahulanand16nov @eguzki @guicassolato curious on your thoughts

eguzki:
It seems a valid proposal. It allows defining also when ratelimit/auth service should be hit and when it should not. 

Both in the RLP and the AuthPolicy, `rules` is a list of rule items. I would like to clarify the difference between empty (0 items existing list) and null (no list defined). in yaml
```yaml
rules: []
```
vs
```yaml
rules: null
```

[...]

PR #206, opened 2022-09-12
default rules from httproute on RateLimitPolicy rules

eguzki:
### what

When the rules are empty (`null` or `empty list`), take the default values from the target resource (HTTPRoute). This implementation goes to the RateLimitPolicy

Partially implements https://github.com/Kuadrant/kuadrant-controller/issues/190

### Verification steps

Run dev env

```
make local-setup
```

Create HTTPRoute

```yaml
kubectl apply -f - <<EOF
---
kind: HTTPRoute
metadata:
  name: toystore
  labels:
    app: toystore
spec:
  parentRefs:
    - name: istio-ingressgateway
      namespace: istio-system
  hostnames: ["*.toystore.com"]
  rules:
    - matches:
        - path:
            type: PathPrefix
            value: "/toy"
          method: GET
        - path:
            type: Exact
            value: "/car"
          method: POST
      backendRefs:
        - name: toystore
          port: 80
EOF
```

Create RateLimitPolicy, with two configurations, one including some rules, another with empty rules

```yaml
kubectl apply -f - <<EOF
---
kind: RateLimitPolicy
metadata:
  name: toystore
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore
  rateLimits:
    - rules:
        - hosts: ["rate-limited.toystore.com"]
      configurations:
        - actions:
            - generic_key:
                descriptor_key: "limited"
                descriptor_value: "1"
      limits:
        - conditions:
            - "limited == 1"
          maxValue: 5
          seconds: 10
          variables: []
    - configurations:
        - actions:
            - generic_key:
                descriptor_key: "limited2"
                descriptor_value: "2"
      limits:
        - conditions:
            - "limited2 == 1"
          maxValue: 2
          seconds: 10
          variables: []
EOF
```
Check the wasm plugin object to see the default rules read from the network resource. One of the rule list is copied directly from the policy, the other is generated from the network resource matching rules.

```yaml
k get wasmplugin kuadrant-istio-ingressgateway -n istio-system -o yaml | yq_pretty 
apiVersion: extensions.istio.io/v1alpha1
kind: WasmPlugin
metadata:
  creationTimestamp: "2022-09-12T12:07:41Z"
  generation: 1
  name: kuadrant-istio-ingressgateway
  namespace: istio-system
  resourceVersion: "2416"
  uid: cca72a57-29e2-4464-861a-a0136ea879f2
spec:
  phase: STATS
  pluginConfig:
    failure_mode_deny: true
    rate_limit_policies:
      - gateway_actions:
          - configurations:
              - actions:
                  - generic_key:
                      descriptor_key: limited
                      descriptor_value: "1"
            rules:
              - hosts:
                  - rate-limited.toystore.com
          - configurations:
              - actions:
                  - generic_key:
                      descriptor_key: limited2
                      descriptor_value: "2"
            rules:
              - hosts:
                  - '*.toystore.com'
                methods:
                  - GET
                paths:
                  - /toy*
              - hosts:
                  - '*.toystore.com'
                methods:
                  - POST
                paths:
                  - /car
        hostnames:
          - '*.toystore.com'
        name: '*.toystore.com'
        rate_limit_domain: istio-system/istio-ingressgateway#*.toystore.com
        upstream_cluster: kuadrant-rate-limiting-service
  selector:
    matchLabels:
      istio: ingressgateway
  url: oci://quay.io/kuadrant/wasm-shim:latest
```

### DIFF
--- a/controllers/apim/ratelimitpolicy_wasm_plugins.go
+++ b/controllers/apim/ratelimitpolicy_wasm_plugins.go
@@ -10,6 +10,7 @@ import (
 	istioclientgoextensionv1alpha1 "istio.io/client-go/pkg/apis/extensions/v1alpha1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	gatewayapiv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
 
 	apimv1alpha1 "github.com/kuadrant/kuadrant-controller/apis/apim/v1alpha1"
 	"github.com/kuadrant/kuadrant-controller/pkg/common"
@@ -150,10 +151,10 @@ func (r *RateLimitPolicyReconciler) wasmPluginConfig(ctx context.Context,
 	if gwRLP != nil {
 		if len(gw.Hostnames()) == 0 {
 			// wildcard domain
-			gatewayActions["*"] = append(gatewayActions["*"], rlptools.GatewayActionsFromRateLimitPolicy(gwRLP)...)
+			gatewayActions["*"] = append(gatewayActions["*"], rlptools.GatewayActionsFromRateLimitPolicy(gwRLP, nil)...)
 		} else {
 			for _, gwHostname := range gw.Hostnames() {
-				gatewayActions[gwHostname] = append(gatewayActions[gwHostname], rlptools.GatewayActionsFromRateLimitPolicy(gwRLP)...)
+				gatewayActions[gwHostname] = append(gatewayActions[gwHostname], rlptools.GatewayActionsFromRateLimitPolicy(gwRLP, nil)...)
 			}
 		}
 	}
@@ -165,7 +166,7 @@ func (r *RateLimitPolicyReconciler) wasmPluginConfig(ctx context.Context,
 		}
 
 		// gateways limits merged with the route level limits
-		mergedGatewayActions := mergeGatewayActions(httpRouteRLP, gwRLP)
+		mergedGatewayActions := mergeGatewayActions(httpRouteRLP, gwRLP, httpRoute)
 		// routeLimits referenced by multiple hostnames
 		for _, hostname := range httpRoute.Spec.Hostnames {
 			gatewayActions[string(hostname)] = append(gatewayActions[string(hostname)], mergedGatewayActions...)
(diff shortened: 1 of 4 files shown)


pkg/rlptools/gateway_utils.go
```diff
@@ -213,3 +214,97 @@ func (g GatewayWrapper) Hostnames() []string {
...
+	if pathMatch.Type == nil {
+		// defaults to path prefix match type
+		suffix = "*"
+	}
+	if *pathMatch.Type == gatewayapiv1alpha2.PathMatchPathPrefix {
+		suffix = "*"
+	}
+	if *pathMatch.Type == gatewayapiv1alpha2.PathMatchExact {
+		suffix = ""
+	}
```
alexsnaps, line 294: Wonder if a `switch` wouldn't be clearer…

alexsnaps (approved):

controllers/apim/ratelimitpolicy_wasm_plugins.go
```diff
@@ -150,10 +151,10 @@ func (r *RateLimitPolicyReconciler) wasmPluginConfig(ctx context.Context,
 	if gwRLP != nil {
 		if len(gw.Hostnames()) == 0 {
 			// wildcard domain
-			gatewayActions["*"] = append(gatewayActions["*"], rlptools.GatewayActionsFromRateLimitPolicy(gwRLP)...)
+			gatewayActions["*"] = append(gatewayActions["*"], rlptools.GatewayActionsFromRateLimitPolicy(gwRLP, nil)...)
```
didierofrivia, line 154: Maybe we could refactor this `GatewayActionsFromRateLimitPolicy` now that it has some added functionality getting the rules from the `HTTPRoute` when a given _RLP_ has no rules... I was thinking that instead of using the second argument as a `route *gatewayapiv1alpha2.HTTPRoute`, we could use the networking object (Gw and Route) and extract the rules from there when no rules are defined and calling the fn the same way for both GW and HTTPRoute... that way maybe we could even get rid of `mergeGatewayActions` and having the logic within `wasmPluginConfig` a bit more descriptive.

pkg/rlptools/gateway_utils.go
```diff
@@ -213,3 +214,97 @@ func (g GatewayWrapper) Hostnames() []string {
...
+
+	val := "/"
+	if pathMatch.Value != nil {
+		val = *pathMatch.Value
+	}
+
+	return []string{val + suffix}
+}
```
didierofrivia, line 302: Yup, `switch` could improve readability:

```go
suffix := ""
switch t := pathMatch.(type){
      case nil:
        sufix = "*"
      case gatewayapiv1alpha2.PathMatchPathPrefix:
        sufix = "*"

...
```

pkg/rlptools/gateway_utils.go
eguzki: thanks for the heads up. Actually, I realized that the code was buggy. If the type is nil, it panics with a `nil pointer dereference`.

pkg/rlptools/gateway_utils.go
eguzki: That code does not work, `gatewayapiv1alpha2.PathMatchPathPrefix` is not a type, but a value. 

The best I could come up with is

```go
// Exact path match                                                                      
suffix := ""                                                                             
if pathMatch.Type == nil || *pathMatch.Type == gatewayapiv1alpha2.PathMatchPathPrefix {  
    // defaults to path prefix match type                                                
    suffix = "*"                                                                         
} 
```

If we ever implement Regexp path matching, this code needs to be extended with an additional if case.

pkg/rlptools/gateway_utils.go
eguzki: I have added a unittest to cover that use case, i.e. when the type is nil.

controllers/apim/ratelimitpolicy_wasm_plugins.go
eguzki: sorry, I do not follow. `GatewayActionsFromRateLimitPolicy` method is getting the rules from the HTTPRoute when the RLP has not rules. 

Can you elaborate more on that? surely, this `wasmPluginConfig` method can be improved for readability and maintainability.

didierofrivia (approved):

PR #210, opened 2022-09-15
rlp validate hosts in rules

eguzki:
### what
Ratelimitpolicy: Validate hosts in rules. Rule hosts should all be subset of targeted network resource hostnames.

Reused implementation to refactor validation in the authpolicy controller.

### verification steps

Setup env
```
make local-setup
```

Create HTTPRoute for `*.toystore.com`

```yaml
kubectl apply -f - <<EOF
---
kind: HTTPRoute
metadata:
  name: toystore
  labels:
    app: toystore
spec:
  parentRefs:
    - name: istio-ingressgateway
      namespace: istio-system
  hostnames: ["*.toystore.com"]
  rules:
    - matches:
        - path:
            type: PathPrefix
            value: "/toy"
          method: GET
      backendRefs:
        - name: toystore
          port: 80
EOF
```

RateLimitPolicy applied for the `toystore` HTTPRoute.

```yaml
kubectl apply -f - <<EOF
---
kind: RateLimitPolicy
metadata:
  name: toystore
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore
  rateLimits:
    - rules:
        - hosts: ["wronghostname.com"]
          methods: ["GET"]
      configurations:
        - actions:
            - generic_key:
                descriptor_key: admin_operation
                descriptor_value: "1"
      limits:
        - conditions:
            - "admin_operation == 1"
          maxValue: 5
          seconds: 10
          variables: []
EOF
``` 
Check RLP status to verify the RLP is not valid

```yaml
k get ratelimitpolicy toystore -o jsonpath='{.status}' | yq e -P
conditions:
  - lastTransitionTime: "2022-09-15T08:27:18Z"
    message: rule host (wronghostname.com) not valid
    reason: ReconcilliationError
    status: "False"
    type: Available
observedGeneration: 1
```
The error msg tells about the hostname having the conflict

Update `wronghostname.com` for `*.ratelimited.toystore.com`
```yaml
kubectl apply -f - <<EOF
---
kind: RateLimitPolicy
metadata:
  name: toystore
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore
  rateLimits:
    - rules:
        - hosts: ["*.ratelimited.toystore.com"]
          methods: ["GET"]
      configurations:
        - actions:
            - generic_key:
                descriptor_key: admin_operation
                descriptor_value: "1"
      limits:
        - conditions:
            - "admin_operation == 1"
          maxValue: 5
          seconds: 10
          variables: []
EOF
``` 

Check RLP status to verify the RLP is now valid

```yaml
k get ratelimitpolicy toystore -o jsonpath='{.status}' | yq e -P
conditions:
  - lastTransitionTime: "2022-09-15T08:32:46Z"
    message: HTTPRoute is ratelimited
    reason: HTTPRouteProtected
    status: "True"
    type: Available
observedGeneration: 2
```

### DIFF
--- a/controllers/apim/authpolicy_controller.go
+++ b/controllers/apim/authpolicy_controller.go
@@ -11,7 +11,6 @@ import (
 	secv1beta1resources "istio.io/client-go/pkg/apis/security/v1beta1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
@@ -22,13 +21,11 @@ import (
 	apimv1alpha1 "github.com/kuadrant/kuadrant-controller/apis/apim/v1alpha1"
 	"github.com/kuadrant/kuadrant-controller/pkg/common"
 	"github.com/kuadrant/kuadrant-controller/pkg/reconcilers"
-	"github.com/kuadrant/kuadrant-controller/pkg/rlptools"
 )
 
 // AuthPolicyReconciler reconciles a AuthPolicy object
 type AuthPolicyReconciler struct {
-	*reconcilers.BaseReconciler
-	Scheme *runtime.Scheme
+	reconcilers.TargetRefReconciler
 }
 
 const authPolicyFinalizer = "authpolicy.kuadrant.io/finalizer"
@@ -66,6 +63,11 @@ func (r *AuthPolicyReconciler) Reconcile(eventCtx context.Context, req ctrl.Requ
 		if err := r.removeAuthSchemes(ctx, &ap); err != nil {
 			return ctrl.Result{}, err
 		}
+
+		if err := r.deleteNetworkResourceBackReference(ctx, &ap); err != nil {
+			return ctrl.Result{}, err
+		}
+
 		controllerutil.RemoveFinalizer(&ap, authPolicyFinalizer)
 		if err := r.UpdateResource(ctx, &ap); client.IgnoreNotFound(err) != nil {
 			return ctrl.Result{}, err
@@ -136,27 +138,9 @@ func (r *AuthPolicyReconciler) reconcileSpec(ctx context.Context, ap *apimv1alph
 }
 
 func (r *AuthPolicyReconciler) enforceHierarchicalConstraints(ctx context.Context, ap *apimv1alpha1.AuthPolicy) error {
-	targetObj, err := r.fetchTargetObject(ctx, ap)
+	targetHostnames, err := r.TargetHostnames(ctx, ap.Spec.TargetRef, ap.Namespace)
 	if err != nil {
 		return err
-	}
-
-	netResourceHosts := make([]string, 0)
-	switch netResource := targetObj.(type) {
-	case *gatewayapiv1alpha2.HTTPRoute:
-		for _, hostname := range netResource.Spec.Hostnames {
-			netResourceHosts = append(netResourceHosts, string(hostname))
-		}
-	case *gatewayapiv1alpha2.Gateway:
-		for idx := range netResource.Spec.Listeners {
-			if netResource.Spec.Listeners[idx].Hostname != nil {
-				netResourceHosts = append(netResourceHosts, string(*netResource.Spec.Listeners[idx].Hostname))
-			}
-		}
-	}
-
-	if len(netResourceHosts) == 0 {
-		netResourceHosts = append(netResourceHosts, string("*"))
 	}
 
 	ruleHosts := make([]string, 0)
@@ -164,11 +148,11 @@ func (r *AuthPolicyReconciler) enforceHierarchicalConstraints(ctx context.Contex
 		ruleHosts = append(ruleHosts, rule.Hosts...)
 	}
 
-	if valid, invalidHost := common.ValidSubdomains(netResourceHosts, ruleHosts); !valid {
+	if valid, invalidHost := common.ValidSubdomains(targetHostnames, ruleHosts); !valid {
 		return fmt.Errorf("rule host (%s) does not follow any hierarchical constraints", invalidHost)
 	}
 
-	if valid, invalidHost := common.ValidSubdomains(netResourceHosts, ap.Spec.AuthScheme.Hosts); !valid {
+	if valid, invalidHost := common.ValidSubdomains(targetHostnames, ap.Spec.AuthScheme.Hosts); !valid {
 		return fmt.Errorf("host defined in authscheme (%s) does not follow any hierarchical constraints", invalidHost)
 	}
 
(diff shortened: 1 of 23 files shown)


controllers/apim/authpolicy_controller.go
```diff
@@ -137,55 +136,42 @@ func (r *AuthPolicyReconciler) reconcileSpec(ctx context.Context, ap *apimv1alph
...
-				trimmedHost := strings.Trim(hostname, "*")
-				constrainingHosts = append(constrainingHosts, trimmedHost)
-			} else {
-				constrainingHosts = append(constrainingHosts, hostname)
+	netResourceHosts := make([]string, 0)
+	switch netResource := targetObj.(type) {
+	case *gatewayapiv1alpha2.HTTPRoute:
+		for _, hostname := range netResource.Spec.Hostnames {
+			netResourceHosts = append(netResourceHosts, string(hostname))
+		}
```
guicassolato, line 149: If only they were of the same kind 😕

```go
netResourceHosts = append(netResourceHosts, netResource.Spec.Hostnames...)
```

controllers/apim/authpolicy_controller.go
```diff
@@ -137,55 +136,42 @@ func (r *AuthPolicyReconciler) reconcileSpec(ctx context.Context, ap *apimv1alph
...
-			}
-		}
-		if !isSubsetHost {
-			return fmt.Errorf("host defined in authscheme (%s) does not follow any hierarchical constraints", host)
-		}
+	if valid, invalidHost := common.ValidSubdomains(netResourceHosts, ruleHosts); !valid {
+		return fmt.Errorf("rule host (%s) does not follow any hierarchical constraints", invalidHost)
 	}
+
+	if valid, invalidHost := common.ValidSubdomains(netResourceHosts, ap.Spec.AuthScheme.Hosts); !valid {
```
guicassolato, line 171: I imagine soon this will go away, right?

controllers/apim/ratelimitpolicy_controller.go
```diff
@@ -380,6 +385,58 @@ func (r *RateLimitPolicyReconciler) gatewayRefListFromHTTPRoute(httpRoute *gatew
...
+		}
+	} else if rlp.IsForHTTPRoute() {
+		netObj, err = r.fetchHTTPRoute(ctx, rlp)
+		if err != nil {
+			// The object should also exist
+			return err
+		}
+	} else {
+		return errors.New("validateRuleHosts: rlp targeting unknown network resource")
+	}
```
guicassolato, line 406: Make sense to move this to another function? Maybe here, maybe a func of `apimv1alpha1.RateLimitPolicy`?

pkg/common/common.go
```diff
@@ -113,3 +113,26 @@ func HostnamesToStrings(hostnames []gatewayapiv1alpha2.Hostname) []string {
...
+		validSubdomain := false
+		for _, domain := range domains {
+			if Name(subdomain).SubsetOf(Name(domain)) {
+				validSubdomain = true
+				break
+			}
+		}
+
+		if !validSubdomain {
+			return false, subdomain
```
guicassolato, line 134: I understand this is a fail-fast strategy. Could it become annoying for long lists of host names containing multiple invalid ones? Perhaps an enhancement for the future.

controllers/apim/ratelimitpolicy_controller.go
```diff
@@ -380,6 +385,58 @@ func (r *RateLimitPolicyReconciler) gatewayRefListFromHTTPRoute(httpRoute *gatew
...
+	ruleHosts := make([]string, 0)
+	for idx := range rlp.Spec.RateLimits {
+		for ruleIdx := range rlp.Spec.RateLimits[idx].Rules {
+			ruleHosts = append(ruleHosts, rlp.Spec.RateLimits[idx].Rules[ruleIdx].Hosts...)
+		}
+	}
+
+	if valid, invalidHost := common.ValidSubdomains(netResourceHosts, ruleHosts); !valid {
+		return fmt.Errorf("rule host (%s) not valid", invalidHost)
+	}
```
guicassolato, line 435: Quite some repetition here compared to `AuthPolicyReconciler.reconcileSpec(...)`. Make sense having a common func that receives the rules (auth or RL) and returns the value for `netResourceHosts`?

controllers/apim/authpolicy_controller.go
eguzki: we could have our Hosts field from the rules of `[]Hostname` type, but we would need to cast anyway

controllers/apim/authpolicy_controller.go
eguzki: that's the plan. For another PR

controllers/apim/ratelimitpolicy_controller.go
eguzki: this will add new changes to the PR, but since you asked :rocket:

pkg/common/common.go
eguzki: Indeed, enhancement for the future.

controllers/apim/ratelimitpolicy_controller.go
eguzki: Makes sense.

controllers/apim/ratelimitpolicy_controller.go
eguzki: Done in https://github.com/Kuadrant/kuadrant-controller/pull/212

controllers/apim/ratelimitpolicy_controller.go
eguzki: Done in https://github.com/Kuadrant/kuadrant-controller/pull/212

eguzki:
all comments addressed @guicassolato

didierofrivia (approved):

guicassolato:
Just caught another issue with the verification steps above, at least if you want to go all the way up to testing the rate-limited host names specified in the RateLimitPolicy (RLP).

The HTTPRoute macthes `GET /toy`. However, the RLP attaches rules to `POST /*`. This means that the RLP cannot be tested.

---

In time, I also deployed the toystore API of course:

```sh
kubectl apply -f https://raw.githubusercontent.com/Kuadrant/kuadrant-controller/main/examples/toystore/toystore.yaml
```

eguzki:
> Just caught another issue with the verification steps above, at least if you want to go all the way up to testing the rate-limited host names specified in the RateLimitPolicy (RLP).
> 
> The HTTPRoute macthes `GET /toy`. However, the RLP attaches rules to `POST /*`. This means that the RLP cannot be tested.
> 
> In time, I also deployed the toystore API of course:
> 
> ```shell
> kubectl apply -f https://raw.githubusercontent.com/Kuadrant/kuadrant-controller/main/examples/toystore/toystore.yaml
> ```

Well, the verification steps only care about the status of the policy CR. That is why the upstream service is not even deployed. 

Regarding your comment, what happens when a POST request hits the gateway? The rate limit filter (implemented by the wasm shim engine) will trigger a call to limitador and rate limit accordingly. Then, the router filter will return 404 because there is no match

I will change the verification steps to GET traffic, just to be easy to test

guicassolato:
> > Just caught another issue with the verification steps above, at least if you want to go all the way up to testing the rate-limited host names specified in the RateLimitPolicy (RLP).
> > The HTTPRoute macthes `GET /toy`. However, the RLP attaches rules to `POST /*`. This means that the RLP cannot be tested.
> > In time, I also deployed the toystore API of course:
> > ```shell
> > kubectl apply -f https://raw.githubusercontent.com/Kuadrant/kuadrant-controller/main/examples/toystore/toystore.yaml
> > ```
> 
> Well, the verification steps only care about the status of the policy CR. That is why the upstream service is not even deployed.
> 
> Regarding your comment, what happens when a POST request hits the gateway? The rate limit filter (implemented by the wasm shim engine) will trigger a call to limitador and rate limit accordingly. Then, the router filter will return 404 because there is no match
> 
> I will change the verification steps to GET traffic, just to be easy to test

Thanks for the explanation, @eguzki! Indeed, all I had to do was adding `GET` method to the rules in the RLP.

eguzki:
added a fix for a issue uncovered when passing RLP integration tests we just merged before this

guicassolato (approved):
For the future:
1. As host names are constraint and validated so Gateway > HTTPRoute > Policy rules, should we explore the same semantics for paths, methods, etc?
2. Raised in #212 in this [comment](https://github.com/Kuadrant/kuadrant-controller/pull/212#pullrequestreview-1110936065), I can see another round of enhancements here to create a proper abstraction for the different target object kinds (HTTPRoute and Gateway), so the caller doesn't need to care about it when getting the rules, the host names, the gateways keys, etc. We can avoid quite a few `if..else`s here and make the code more readable and maintainable to later support other kinds of target objects as well (Namespaces, Backends, OpenShift Routes, etc).


For this iteration, LGTM. I tried locally including with the AuthPolicy and it worked beautifully. Nice work, @eguzki!

PR #212, opened 2022-09-15
Refactor: dry enhancements

eguzki:
### what

This is a refactor of the ratelimitpolicy and authpolicy controller to use common code that was repeated.

For that, a new base reconciler is created called `TargetRefReconciler` which is meant for controllers reconciling CRD's having a [targetRef from the Gateway API policy attachment](https://gateway-api.sigs.k8s.io/references/policy-attachment/#target-reference-api). Also, a new file in the package `commons` called `gatewayapi_utils.go` with common functionality regarding gateway API resources.

* Remove back reference of the authpolicy in the target object when the authpolicy is being deleted.
* Unittests of the common code

### DIFF
--- a/main.go
+++ b/main.go
@@ -121,8 +121,9 @@ func main() {
 	)
 
 	if err = (&apimcontrollers.RateLimitPolicyReconciler{
-		BaseReconciler: rateLimitPolicyBaseReconciler,
-		Scheme:         mgr.GetScheme(),
+		TargetRefReconciler: reconcilers.TargetRefReconciler{
+			BaseReconciler: rateLimitPolicyBaseReconciler,
+		},
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "RateLimitPolicy")
 		os.Exit(1)
@@ -135,8 +136,9 @@ func main() {
 	)
 
 	if err = (&apimcontrollers.AuthPolicyReconciler{
-		BaseReconciler: authPolicyBaseReconciler,
-		Scheme:         mgr.GetScheme(),
+		TargetRefReconciler: reconcilers.TargetRefReconciler{
+			BaseReconciler: authPolicyBaseReconciler,
+		},
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "AuthPolicy")
 		os.Exit(1)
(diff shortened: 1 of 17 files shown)


main.go
```diff
@@ -121,8 +121,9 @@ func main() {
 	)
 
 	if err = (&apimcontrollers.RateLimitPolicyReconciler{
-		BaseReconciler: rateLimitPolicyBaseReconciler,
-		Scheme:         mgr.GetScheme(),
+		TargetRefReconciler: reconcilers.TargetRefReconciler{
+			BaseReconciler: rateLimitPolicyBaseReconciler,
+		},
```
guicassolato, line 126:
```suggestion
		TargetRefReconciler: reconcilers.NewTargetRefReconciler(rateLimitPolicyBaseReconciler),
```

guicassolato (commented):
This is good refactoring already. Perhaps one iteration further and we could maybe try to reduce the number of `if..else` and `switch` statements now checking the type of the target ref object. WDYT about an interface to abstract any target ref object for the functions we need here (i.e. `GetRules()`, `GetHostNames()`, `GetKind()`, `GetTargetObject()`, `ValidSubdomains()`, `GetGatewayKeys()`, etc), with proper implementations for `Gateway` and for `HTTPRoute`?

eguzki:
> This is good refactoring already. Perhaps one iteration further and we could maybe try to reduce the number of `if..else` and `switch` statements now checking the type of the target ref object. WDYT about an interface to abstract any target ref object for the functions we need here (i.e. `GetRules()`, `GetHostNames()`, `GetKind()`, `GetTargetObject()`, `ValidSubdomains()`, `GetGatewayKeys()`, etc), with proper implementations for `Gateway` and for `HTTPRoute`?

yeah, there is plenty of room for improvement. I will happily review that abstraction ;)

main.go
eguzki: if there was anything to be done by the builder, then it would make sense. Currently, it is just a cosmetic change

guicassolato: I see it as abstraction and a good practice rather than just cosmetic, but I guess the same argument could be used to propose also `NewRateLimitPolicyReconciler()` and `NewAuthPolicyReconciler()`. Not critical anyway.

guicassolato (approved):
> I will happily review that abstraction ;)

Let's do it in another PR then?

didierofrivia (approved):

PR #213, opened 2022-09-19
rlp e2e test

### LINKED ISSUE #179: Add E2E tests to cover RateLimitPolicy API

eguzki:
### what

e2e tests to cover RateLimitPolicy

Fixes #179 

It is a happy path testing and does not test all the functionality, but since we are already thinking on a breaking change of the API (introducing defaults/overrides), better something than nothing to rely on.

### Verification steps

GH Action checks should pass :)

### DIFF
--- a/controllers/apim/ratelimitpolicy_controller_test.go
+++ b/controllers/apim/ratelimitpolicy_controller_test.go
@@ -1,0 +1,269 @@
+//go:build integration
+
+package apim
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	istioclientgoextensionv1alpha1 "istio.io/client-go/pkg/apis/extensions/v1alpha1"
+	istioclientnetworkingv1alpha3 "istio.io/client-go/pkg/apis/networking/v1alpha3"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	gatewayapiv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
+
+	apimv1alpha1 "github.com/kuadrant/kuadrant-controller/apis/apim/v1alpha1"
+	"github.com/kuadrant/kuadrant-controller/pkg/common"
+	"github.com/kuadrant/kuadrant-controller/pkg/rlptools"
+	limitadorv1alpha1 "github.com/kuadrant/limitador-operator/api/v1alpha1"
+)
+
+func testBuildBasicGateway(gwName, ns string) *gatewayapiv1alpha2.Gateway {
+	return &gatewayapiv1alpha2.Gateway{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Gateway",
+			APIVersion: gatewayapiv1alpha2.GroupVersion.String(),
+		},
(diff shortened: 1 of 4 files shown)


controllers/apim/ratelimitpolicy_controller_test.go
```diff
@@ -0,0 +1,269 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	gatewayapiv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
+
+	apimv1alpha1 "github.com/kuadrant/kuadrant-controller/apis/apim/v1alpha1"
+	"github.com/kuadrant/kuadrant-controller/pkg/common"
+	"github.com/kuadrant/kuadrant-controller/pkg/rlptools"
+	limitadorv1alpha1 "github.com/kuadrant/limitador-operator/api/v1alpha1"
+)
+
+func testBuildBasicGateway(gwName, ns string) *gatewayapiv1alpha2.Gateway {
```
didierofrivia, line 26: Maybe we could name this one, and `testBuildBasicHttpRoute`, `testBuildBasicRoutePolicy`, etc... `fake` or even `mock`, since those are not testing but acting as a _double_ . I.E: `mockBasicHttpRoute`

controllers/apim/ratelimitpolicy_controller_test.go
```diff
@@ -0,0 +1,269 @@
...
+	AfterEach(DeleteNamespaceCallback(&testNamespace))
+
+	Context("Basic: RLP targeting HTTPRoute", func() {
+		It("check created resources", func() {
+
+			routeName := "toystore-route"
+			gwName := "toystore-gw"
+			rlpName := "toystore-rlp"
+
+			gateway := testBuildBasicGateway(gwName, testNamespace)
```
didierofrivia, line 154: Maybe there's a way to have _Factories_ for these mocked instances, so we could reuse them in other tests and build them with specific params? Something similar to "FactoryGirl" in Ruby maybe ?

controllers/apim/suite_test.go
```diff
@@ -80,6 +82,18 @@ var _ = BeforeSuite(func() {
 	err = istiosecurityv1beta1.AddToScheme(scheme.Scheme)
 	Expect(err).NotTo(HaveOccurred())
 
+	err = limitadorv1alpha1.AddToScheme(scheme.Scheme)
+	Expect(err).NotTo(HaveOccurred())
+
+	err = limitadorv1alpha1.AddToScheme(scheme.Scheme)
+	Expect(err).NotTo(HaveOccurred())
```
didierofrivia, line 89: Is this duplication intentional?

controllers/apim/ratelimitpolicy_controller_test.go
eguzki: well, it is not a mock. It is just a builder method. I added `test` prefix to avoid collision with any other name from the same package. 

In go, regular test starts with `TestXXXX` so there is no issue with that.

controllers/apim/suite_test.go
eguzki: good catch. I will remove

controllers/apim/ratelimitpolicy_controller_test.go
eguzki: they can be reused from all the tests in the same package. 

But, yeah.. we can do much better. maybe in another PR?

controllers/apim/suite_test.go
eguzki: done

didierofrivia (approved):

